PR #21, opened 2020-10-06
Improve tests

oysteoh:


ManInFez:
Maybe we could create a `response-model`, which would contain all the data which we would need, and the axis information would be in `response.axis`? The data_loader would then create a response model?

oysteoh:
> Maybe we could create a `response-model`, which would contain all the data which we would need, and the axis information would be in `response.axis`? The data_loader would then create a response model?

Done! I hope... ? 👍  If this is what you meant? Currently it's looking a bit weird since the controller pulls out realizations and observations as well from the response - but i will take this bit by bit while discussing with you !

ertviz/controllers/time_series_controller.py
```diff
@@ -113,13 +107,14 @@ def _update_graph(value):
...
+
+        x_axis = response.axis
+        realizations = _get_realizations_data(response.realizations, x_axis)
         observations = []
 
-        for obs in response.get("observations", []):
-            observations += _get_observation_data(obs["data"], x_axis)
+        for obs in response.observations:
+            observations += _get_observation_data(obs, x_axis)
         print(len(observations))
```
ManInFez, line 118: remove print

ManInFez (approved):
minor comment

PR #30, opened 2020-10-21
Ensemble model

xjules:
Involves:
- a new api for data access 
- controllers creates plots from pandas objects instead ert objects.
- webgl support for some views, which dramatically improves performance 
- basic statistics plot for responses 
- tests


ertviz/controllers/response_controller.py
```diff
@@ -51,33 +92,35 @@ def _get_observation_data(observation, x_axis):
...
+        mode="lines",
         line=dict(color="red", dash="dash"),
         marker=None,
     )
     return [observation_data, lower_std_data, upper_std_data]
 
 
-def _create_response_model(url):
-    response = get_response(url)
+def _create_response_model(response, plot_type, selected_realizations):
```
ManInFez, line 102: Can we rename this to `_create_response_plot` and maybe rename EnsemblePlotModel to ResponsePlotModel? Maybe we could have a little brainstorm on naming

ertviz/controllers/response_controller.py
```diff
@@ -1,83 +1,113 @@
...
-from ertviz.data_loader import get_ensemble
+from ertviz.data_loader import get_ensemble_url
 from ertviz.ert_client import get_response
-from ertviz.models import EnsemblePlotModel, PlotModel
+from ertviz.models import ResponsePlotModel, PlotModel, EnsembleModel
 from ertviz.controllers import parse_url_query
 
 
-def _get_realizations_data(realizations, x_axis):
+obs_color = "rgb(176, 28, 52)"
```
dotfloat, line 12: Should we put this in a stylesheet?

ertviz/models/plot_model.py
```diff
@@ -1,12 +1,16 @@
+import plotly.graph_objects as go
+
+
 class PlotModel:
-    def __init__(self, x_axis, y_axis, text, name, mode, line, marker):
+    def __init__(self, x_axis, y_axis, text, name, mode, line, marker, error_y=None):
```
dotfloat, line 5: Consider using kwargs or making setters for this in the future. Too many arguments.

ertviz/models/plot_model.py
```diff
@@ -26,21 +31,21 @@ def repr(self):
...
+        return go.Scattergl(repr_dict)
 
     @property
     def name(self):
         return self._name
 
     @property
     def display_name(self):
-        if self._name is int:
+        if type(self._name) is int:
```
dotfloat, line 42: `isinstance(self._name, int)`

tests/data/snake_oil_data.py
```diff
@@ -53,39 +53,90 @@ def content(self):
                 },
                 "ref_url": "http://127.0.0.1:5000/ensembles/1/parameters/1",
             },
+            # {
```
dotfloat, line 56: Do something with these comments.

ertviz/controllers/response_controller.py
xjules: Thought about it and It might be a good idea :) It should include most of the properties then.

ertviz/models/plot_model.py
xjules: Agree!

ertviz/controllers/response_controller.py
xjules: Let's create an issue about a dedicated `.css` file once this'd been merged.

dotfloat (approved):
:tada:

PR #32, opened 2020-10-26
Lazy loading for responses

ManInFez:
Increase lazy loading for responses to prevent fetching schemas before its needed


ertviz/ert_client/__init__.py
```diff
@@ -1,8 +1,8 @@
 from ertviz.models import Response, EnsembleModel
 
 
-def get_response(ref_url):
-    return Response(ref_url=ref_url)
+def get_response(name, ref_url):
```
xjules, line 4: Is this function even used somewhere?

ManInFez: Maybe not

xjules: Then we can remove it (from tests too I guess).

ManInFez: done!

ertviz/models/response.py
```diff
@@ -4,32 +4,38 @@
...
+
+    @property
+    def ensemble_id(self):
+        if not self._schema:
+            self._schema = get_schema(api_url=self._ref_url)
+        return self._schema["ensemble_id"]
 
     @property
     def axis(self):
+        if not self._schema:
```
xjules, line 26: this's quite repetitive, could we make a function e.g. `_update_schema` 
```python
def _update_schema
   if not self._schema:
          self._schema = get_schema(api_url=self._ref_url)
```

ManInFez: :+1:

xjules (approved):
I think it is good to go now! Bra job!

PR #33, opened 2020-10-28
Add kernel density estimate plot to parameter plot

### LINKED ISSUE #31: KDE plot for parameters
Besides histogram, we should provide KDE plot, something like this:
image

xjules:
Here check boxes for histogram and kde are added bellow the parameter selection listbox. Additionally, to speed loading time, lazy loading option is used for parameters.
Resolves #31


ManInFez (approved):
LGTM :+1:

PR #34, opened 2020-10-30
Initialize parameters plot through pandas DataFrame

xjules:
To separate schema / rest api from plots, also parameters plot need to be initiated from DataFrame. This also simplifies test implementation for plots.
Resoloves #39

### DIFF
--- a/ertviz/controllers/parameter_controller.py
+++ b/ertviz/controllers/parameter_controller.py
@@ -1,7 +1,4 @@
-import re
-from tests import data
 import dash
-import json
 from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output, State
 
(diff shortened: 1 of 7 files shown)


ertviz/controllers/parameter_controller.py
```diff
@@ -1,11 +1,12 @@
 import re
+from tests import data
```
ManInFez, line 2: should this be here?

ManInFez (approved):
just a small import to remove :+1:

ertviz/controllers/parameter_controller.py
xjules: yes, that's a legacy from before. Also `import json` can go away

PR #35, opened 2020-11-02
Comparison plugin

### LINKED ISSUE #16: Create tree view for Ensemble relationships
An ensamble may have a parent and sevaral children. A tree view would be a good way of obtaining an overview of how these are connected.

ManInFez:
Add plugin for showing multiple responses in one plot

resolves equinor/ert#769
resolves #16

### DIFF
--- a/ertviz/controllers/ensemble_selector_controller.py
+++ b/ertviz/controllers/ensemble_selector_controller.py
@@ -1,0 +1,110 @@
+import json
+import random
+import dash
+from dash.dependencies import Input, Output, State
+from dash.exceptions import PreventUpdate
+from ertviz.data_loader import get_ensembles
+from ertviz.models import EnsembleModel
+
+
+def _construct_graph(ensembles):
+
+    roots = [
+        ensemble
+        for ensemble_id, ensemble in ensembles.items()
+        if ensemble.parent == None
+    ]
+
+    datas = []
+    queue = []
+    for index, ensemble in enumerate(roots):
+        queue.append(ensemble)
+
+    def _construct_node(ensembl):
+        return {
+            "data": {
+                "id": ensemble.id,
+                "label": ensemble._name,
+            }
+        }
+
+    def _construct_edge(ensemble_src, ensemble_target):
+        return {
+            "data": {
+                "source": ensemble_src.id,
+                "target": ensemble_target.id,
+            }
+        }
+
+    while queue:
+        ensemble = queue.pop(-1)
+        datas.append(_construct_node(ensemble))
+        nr_of_children = len(ensemble.children)
+        for index, child in enumerate(ensemble.children):
+            datas.append(_construct_edge(ensemble, child))
+            queue.append(child)
+    return datas
+
+
+def ensemble_selector_controller(parent, app):
+    @app.callback(
+        Output(parent.uuid("ensemble-selector"), "layout"),
+        [Input("dropdown-layout", "value")],
+    )
+    def update_cytoscape_layout(layout):
+        return {"name": layout}
+
+    @app.callback(
+        Output(parent.uuid("ensemble-selector"), "elements"),
+        [
+            Input("url", "pathname"),
+            Input(parent.uuid("ensemble-selection-store"), "data"),
+        ],
+        [
+            State(parent.uuid("ensemble-selector"), "elements"),
+        ],
+    )
+    def update_ensemble_selector(query, selected_ensembles, elements):
+        ctx = dash.callback_context
+        select_update = ctx.triggered[0]["prop_id"].split(".")[0] == parent.uuid(
+            "ensemble-selection-store"
+        )
+        if select_update:
+            datas = elements
+            for element in datas:
+                element["data"].update(
+                    selected_ensembles.get(element["data"]["id"], {})
+                )
+        else:
+            ensemble_dict = get_ensembles()
+            for ensemble_schema in ensemble_dict:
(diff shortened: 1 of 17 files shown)


ertviz/controllers/ensemble_selector_controller.py
```diff
@@ -0,0 +1,88 @@
...
+        Output(parent.uuid("ensemble-selector"), "layout"),
+        [Input("dropdown-layout", "value")],
+    )
+    def update_cytoscape_layout(layout):
+        return {"name": layout}
+
+    @app.callback(
+        Output(parent.uuid("ensemble-selector"), "elements"), [Input("url", "pathname")]
+    )
+    def update_ensemble_selector(query):
```
xjules, line 20: where's the query param (url) actually used?

ertviz/controllers/ensemble_selector_controller.py
```diff
@@ -0,0 +1,88 @@
...
+
+        def _construct_edge(ensemble_src, ensemble_target):
+            return {
+                "data": {
+                    "source": ensemble_src.id,
+                    "target": ensemble_target.id,
+                }
+            }
+
+        while queue:
```
xjules, line 59: I would encapsulate this to a function, which would take pandas df instead - to alleviate testing. Similarly to `def _get_realizations_plots` in `response_controller.py`

ManInFez: This does not fit into a pandas frame, as it is relational data.

xjules: I'll look at it again ... but the idea was to get the unit tests for the graph into `test_controller.py` What do you think

xjules:
It looks very nice! I thought, shall we remove completely ensemble viewer and replace it with comparison viewer then?

ertviz/views/__init__.py
```diff
@@ -1,12 +1,31 @@
...
+            ),
+        ]
+    )
 
 
 def response_view(parent):
     return [
         html.Div(
             [
+                _graph_type_div(
```
xjules, line 24: Once changing this, can we move this `_graph_type_div` under the response selector? ie.  
```python
html.Div(
                    [
                        html.H5("Response"),
                        dcc.Dropdown(id=parent.uuid("response-selector")),
                        _graph_type_div(
                            parent=parent,
                            name="graph_type",
                            keys=["Function plot", "Statistics"],
                        ),
                    ],
```

ManInFez:
> It looks very nice! I thought, shall we remove completely ensemble viewer and replace it with comparison viewer then?

Maybe, I think it is up for debate. we might want to have a single ensemble details view...

ertviz/controllers/ensemble_selector_controller.py
ManInFez: A test would probably be good

ertviz/controllers/ensemble_selector_controller.py
ManInFez: it is not actually used, but url input is used as a trigger to update the graph. Might be a better way for this..

ertviz/models/ensemble_model.py
```diff
@@ -39,18 +39,19 @@ def get_parameter_models(parameters_schema):
 
 class EnsembleModel:
     def __init__(self, ref_url):
-        schema = get_schema(api_url=ref_url)
-        self._name = schema["name"]
+        self._schema = get_schema(api_url=ref_url)
+        self._name = self._schema["name"]
         self._id = ref_url  # ref_url
```
oysteoh, line 44: seems like an unnecessary comment?

ertviz/views/ensemble_selector_view.py
```diff
@@ -0,0 +1,55 @@
...
+            [
+                cyto.Cytoscape(
+                    id=parent.uuid("ensemble-selector"),
+                    layout={"name": "grid"},
+                    style=cyto_style,
+                    stylesheet=[
+                        {"selector": "node", "style": {"label": "data(label)"}},
+                        {
+                            "selector": "edge",
+                            "style": {"line-color": "rgba(55,55,55,0.8)"},
```
oysteoh, line 41: would it make sense having all the colors and such in a separate config file / file where we load those from?

ManInFez: I think we should at some point move to a style sheet to provide all stylings. I vote we keep it in for now and I will make an issue of using a CSS for styling ertviz :)

ertviz/models/ensemble_model.py
ManInFez: It was meant to be a bit more descriptive. For now the unique identifier for things are the address, for it to be fetched from the api. It might be that at one point we are using a more structured http api and then maybe it makes more sense to pass in the id in the schema, and construct http calls based on those ids. Removing the comment for now.

oysteoh:
> > It looks very nice! I thought, shall we remove completely ensemble viewer and replace it with comparison viewer then?
> 
> Maybe, I think it is up for debate. we might want to have a single ensemble details view...

I would think it is some wanted feature in the future displaying some metadata regarding the ensemble in it's own detailed view. Now when we have the rdb it's easier to extract relevant data.

oysteoh (approved):
LGTM 👍

ertviz/controllers/ensemble_selector_controller.py
xjules: So you did the test, nice!

PR #44, opened 2020-11-12
Use styles from assets

### LINKED ISSUE #37: Use CSS for ertviz
There is currently many views in ertviz with styles defined inline. The styles should move out of the code and into a CSS file. This CSS file should reflect the style of equinor pages, see https://webviz-subsurface-example.azurewebsites.net/ for an example

ManInFez:
resolves #37


oysteoh (approved):

sondreso:
Screenshot? :smile:

Issue #36, opened 2020-11-04
Write dash tests


oysteoh:

oysteoh:
ongoing work in the different views.

PR #45, opened 2020-11-12
Testing view ensemble_overview

oysteoh:
Relates to #36

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,14 @@
-import sys
-import pytest
 import pandas as pd
-from tests.data.snake_oil_data import ensembles_response
 from requests import HTTPError
-from unittest.mock import Mock, MagicMock
+
+from tests.data.snake_oil_data import ensembles_response
+from selenium.webdriver.chrome.options import Options
+
+
+def pytest_setup_options():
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -1,9 +1,15 @@
...
-from unittest.mock import Mock, MagicMock
+
+from tests.data.snake_oil_data import ensembles_response
+from selenium.webdriver.chrome.options import Options
+
+
+def pytest_setup_options():
+    options = Options()
+    options.add_argument("--headless")
+    options.add_argument("--disable-gpu")
```
ManInFez, line 11: Some of our components are hardware accelerated, will disable-gpu affect testing those?

oysteoh: hmm... i guess we can remove that and only run with `--headless` ? I think that should be sufficient! I'll change and check if the tests are running 👍

ManInFez (approved):
Good first step in testing dash! We should strive to test all our views :)

oysteoh:
> Good first step in testing dash! We should strive to test all our views :)

Yes, i suggest we are not closing the issue before all views are tested! 👍

PR #51, opened 2020-11-17
Adjust to changes in ERTAPI

oysteoh:
Changes made in the datamodel in ERTAPI - this PR depends on https://github.com/equinor/ert/pull/1170

### DIFF
--- a/ertviz/models/realization.py
+++ b/ertviz/models/realization.py
@@ -1,18 +1,4 @@
-from ertviz.data_loader import get_data
-
-
 class Realization:
     def __init__(self, realization_schema):
-        self._name = realization_schema["name"]
-        self._data_url = realization_schema["data_url"]
-        self._data = None
-
-    @property
-    def data(self):
-        if self._data is None:
-            self._data = get_data(self._data_url)
-        return self._data
-
-    @property
-    def name(self):
-        return self._name
+        self.name = realization_schema["name"]
+        self.data = realization_schema["data"]
--- a/tests/data/snake_oil_data.py
+++ b/tests/data/snake_oil_data.py
@@ -108,21 +108,21 @@ ensembles_response = {
         "key": "BPR_138_PERSISTENCE",
         "parameter_realizations": [
             {
-                "data_url": "http://127.0.0.1:5000/data/33",
+                "data": "http://127.0.0.1:5000/data/33",
                 "name": 0,
                 "realization": {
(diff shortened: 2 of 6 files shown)


oysteoh:
By checking out and using the referenced PR the reviewer should now be able to run this PR without issues visualizing the data from ERT

ertviz/models/realization.py
```diff
@@ -1,18 +1,4 @@
...
-    def data(self):
-        if self._data is None:
-            self._data = get_data(self._data_url)
-        return self._data
-
-    @property
-    def name(self):
-        return self._name
+        self.name = realization_schema["name"]
+        self.data = realization_schema["data"]
```
ManInFez, line 4: if the `realization_schema["data"]` can be both an array of data and a url, how will the actual data be retrieved when it is a URL?

oysteoh: 🤔  you probably have a point, but i don't get it...? can it be both? I think we are only sending "data" now along with the values?

ManInFez: in the test data set it seems to be urls

oysteoh: hmm.... where exactly? I think i've changed all - but i'm might be mistaken here?

oysteoh: forget it - now i see! 👍  will fix!

tests/data/snake_oil_data.py
```diff
@@ -140,41 +140,41 @@ def content(self):
...
+        "axis": {"data": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]},
         "ensemble_id": "1",
         "name": "SNAKE_OIL_GPR_DIFF",
         "realizations": [
             {
-                "data_url": "http://127.0.0.1:5000/data/284",
+                "data": "http://127.0.0.1:5000/data/284",
                 "name": 0,
                 "ref_url": "http://127.0.0.1:5000/ensembles/1/realizations/0",
                 "summarized_misfits": {},
```
ManInFez, line 180: data is here a URL

oysteoh: Yes i forgot to change the values on the data not being used ... fixed for those used in the test... Now i replaced all the urls with just dummy data - i'm not sure how to approach this with a sane test data set - but that we can discuss later 👍

ManInFez (approved):
LGTM!

PR #53, opened 2020-11-18
clean up various views and plugins

ManInFez:


ertviz/plugins/_ensemble_overview.py
```diff
@@ -2,32 +2,26 @@
 import dash_core_components as dcc
 from webviz_config import WebvizPluginABC
 from ertviz.data_loader import get_ensembles
```
oysteoh, line 4: get_ensembles is unused

ertviz/views/paremeter_view.py
```diff
@@ -8,58 +8,60 @@
 import ertviz.assets as assets
```
oysteoh, line 8: a lot of unused imports in this file

oysteoh (approved):
LGTM if you clean up imports along with fixing the error I pointed out 
```
    def _construct_node(ensembl):
        return {
            "data": {
                "id": ensemble.id,
                "label": ensemble._name,
            }
        }
```

PR #58, opened 2020-12-02
Add multiple response viewer plugin

### LINKED ISSUE #52: Visualize comparison between responses

ManInFez:
resolves #52

### DIFF
--- a/ertviz/assets/ert-style.css
+++ b/ertviz/assets/ert-style.css
@@ -18,16 +18,16 @@
 }
 
 .ert-ensemble-selector {
-    width: auto !important;
     height: 200px !important;
     border-color: rgba(181, 223, 223, 0.5);
     border-style: solid;
     border-width: 5px;
-
+    flex: 1;
 }
 .ert-ensemble-selector-container {
-    display: block;
-
+    display: flex;
+    flex-direction: initial;
+    width: 100%
 }
 .ert-graph-container {
     display: table;
@@ -46,4 +46,8 @@
     vertical-align: middle;
     height: 5in;
     width: 100%;
+}
+.ert-graph-multi{
+    width: 50% ;
+    float: left;
 }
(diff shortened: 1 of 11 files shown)


ertviz/assets/webviz-config.yml
```diff
@@ -9,5 +9,10 @@ pages:
...
  - title: Comparison Viewer
    content:
-    - EnsembleComparison
+    - Ensemble Comparison
     - EnsembleComparison:
+
+ - title: Response Comparison Viewer
+   content:
+    - Response Comparison
+    - ResponseComparison:
```
oysteoh, line 18: newline

ertviz/assets/ert-style.css
```diff
@@ -47,3 +47,7 @@
     height: 5in;
     width: 100%;
 }
+.ert-graph-multi{
+    width: 50% ;
+    float: left;
+}
```
oysteoh, line 53: newline

ertviz/models/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 
 def _convertdate(dstring):
-    return datetime.strptime(dstring, "%Y-%m-%d %H:%M:%S")
+    return datetime.strptime(dstring, "%a, %d %b %Y %H:%M:%S GMT")
```
oysteoh, line 5: should we not use ISO8601 for the time representation? 
https://www.w3.org/TR/NOTE-datetime

i.e something like 1997-07-16T19:20:30.45+01:00 and then let the browser ( wherever located ) handle how to represent this in the html?

ManInFez: This is the format currently coming from ert-storage, so it would have to be fixed there

oysteoh: I'll create an issue on it for art-storage - at least to look at how it is sent. Because this is not the best way to send it right?

ManInFez: I think it more sane to send them the timestamps as integers? @dotfloat, any opinions?

oysteoh (approved):
🚀

PR #65, opened 2020-12-14
Make data loading FastAPI compatible

### LINKED ISSUE #64: Use refactored ERT Storage API (FastAPI)
The API will mutate, however we can begin working on it now to see where the holes are.

Relevant PR in ERT: https://github.com/equinor/ert/pull/1105

ManInFez:
resolves #64 

depends on equinor/ert#1105

### DIFF
--- a/tests/data/snake_oil_data.py
+++ b/tests/data/snake_oil_data.py
@@ -14,12 +14,12 @@ ensembles_response = {
                 "children": [
                     {
                         "name": "default_smoother_update",
-                        "ref_url": "http://127.0.0.1:5000/ensembles/2",
+                        "id": 2,
                     }
                 ],
                 "name": "default",
                 "parent": {},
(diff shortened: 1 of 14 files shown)


oysteoh (approved):
ERT 1105 is merged

tests/data/snake_oil_data.py
```diff
@@ -176,7 +170,7 @@ def content(self):
             {
                 "data": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
                 "name": 0,
-                "ref_url": "http://127.0.0.1:5000/ensembles/1/realizations/0",
+                "id": "http://127.0.0.1:5000/ensembles/1/realizations/0",
```
sondreso, line 173: Shouldn't the URL also change here? :thinking:

PR #91, opened 2021-01-06
Add InputBox for user defined number of bins

### LINKED ISSUE #74: Allow for custom bin-size in the histogram

xjules:
Resolves #74

- Instead of `bin_size` it makes sense to specify the number of bins.
- The default `bin_size` was adjusted to match the old ert-gui.


ertviz/models/plot_model.py
```diff
@@ -146,10 +155,9 @@ def repr(self):
                 ]
             )
 
-        bin_count = int(math.ceil(math.sqrt(len(data[0]))))
         _max = max(map(max, data))
         _min = min(map(min, data))
-        bin_size = float((_max - _min) / bin_count)
+        bin_size = float((_max - _min) / self.bin_count)
```
oysteoh, line 160: self._bin_count i think it is called?

ertviz/views/paremeter_view.py
```diff
@@ -47,6 +47,14 @@ def parameter_view(parent):
                             ],
                             value=["hist", "kde"],
                         ),
+                        html.Label("#Bins:"),
```
oysteoh, line 50: If available space - would it be nicer with `number of bins` ?

ertviz/models/plot_model.py
xjules: `self.bin_count` is a property, which recomputes `bin_count` if necessary

ertviz/views/paremeter_view.py
xjules: Wanted to save space, will check if the plot shrinks

ertviz/models/plot_model.py
oysteoh: yes - of course... my mistake 😬

ertviz/views/paremeter_view.py
xjules: The plot didn't shrink, so changed.

oysteoh (approved):
Please squash before merging! 🚀

PR #92, opened 2021-01-07
Add Realization num to histogram rug plot hovering

### LINKED ISSUE #83: Add realisation number to the hovering of the data in parameter plot

autt:
Resolves #83

### DIFF
--- a/ertviz/models/plot_model.py
+++ b/ertviz/models/plot_model.py
@@ -134,10 +134,17 @@ class MultiHistogramPlotModel:
         colors = []
         data = []
         names = []
+        realization_nums = []
         for response_name in self._data_df_dict:
             data.append(list(self._data_df_dict[response_name].values.flatten()))
             colors.append(self._colors[response_name])
             names.append(response_name)
+            realization_nums.append(
+                [
+                    f"Realization {num}"
+                    for num in self._data_df_dict[response_name].columns
+                ]
+            )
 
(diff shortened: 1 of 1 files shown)


ertviz/models/plot_model.py
```diff
@@ -134,10 +134,17 @@ def repr(self):
         colors = []
         data = []
         names = []
+        realization_num = []
```
xjules, line 137: Nice, was wondering if you could change the variable name to plural, eg. realization_nums, for the sake of clarity that we talk many.

autt: Sure!

ertviz/models/plot_model.py
```diff
@@ -134,10 +134,17 @@ def repr(self):
         colors = []
         data = []
         names = []
+        realization_nums = []
         for response_name in self._data_df_dict:
             data.append(list(self._data_df_dict[response_name].values.flatten()))
             colors.append(self._colors[response_name])
             names.append(response_name)
+            realization_nums.append(
+                list(
```
dotfloat, line 143: `list(foo for foo in bar)` is the same as `[foo for foo in bar]`. The first is a generator comprehension given to the `list` function to make it into a `list`, the second is a list comprehension that does it for you and doesn't require an extra indent.

xjules (approved):
Good job, LGTM!

PR #93, opened 2021-01-13
Prior plotting

ManInFez:
Using scipy stats module to plot theoretical priors, not all are complete, as the sampling is not always folloing the theoretical mathematical function it says it does.

resolves https://github.com/equinor/ert/issues/772

### DIFF
--- a/ertviz/controllers/ensemble_selector_controller.py
+++ b/ertviz/controllers/ensemble_selector_controller.py
@@ -2,7 +2,7 @@ import dash
 from dash.dependencies import Input, Output, State
 from dash.exceptions import PreventUpdate
 from ertviz.data_loader import get_ensembles
-from ertviz.models import EnsembleModel
+from ertviz.models import EnsembleModel, load_ensemble
 import ertviz.assets as assets
 
 
@@ -78,11 +78,7 @@ def ensemble_selector_controller(parent, app):
             ensemble_dict = get_ensembles(project_id=parent.project_identifier)
             for ensemble_schema in ensemble_dict:
                 ensemble_id = ensemble_schema["id"]
-                if ensemble_id not in parent.ensembles:
-                    parent.ensembles[ensemble_id] = EnsembleModel(
-                        ensemble_id=int(ensemble_id),
-                        project_id=parent.project_identifier,
-                    )
+                load_ensemble(parent, ensemble_id)
             datas = _construct_graph(parent.ensembles)
         return datas
 
--- a/ertviz/controllers/multi_parameter_controller.py
+++ b/ertviz/controllers/multi_parameter_controller.py
@@ -1,7 +1,7 @@
 import dash
 from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output, State
-from ertviz.models import EnsembleModel, MultiHistogramPlotModel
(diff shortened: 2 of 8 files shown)


ertviz/controllers/ensemble_selector_controller.py
```diff
@@ -78,11 +78,7 @@ def update_ensemble_selector(query, selected_ensembles, elements):
             ensemble_dict = get_ensembles(project_id=parent.project_identifier)
             for ensemble_schema in ensemble_dict:
                 ensemble_id = ensemble_schema["id"]
-                if ensemble_id not in parent.ensembles:
-                    parent.ensembles[ensemble_id] = EnsembleModel(
-                        ensemble_id=int(ensemble_id),
-                        project_id=parent.project_identifier,
-                    )
+                get_ensemble(parent, ensemble_id)
```
xjules, line 81: Shouldn't this provide a handle to ensemble?

ManInFez: maybe the name is wrong, as it not only returns the ensemble, but loads it...

ManInFez: renamed function to load ensemble :)

ertviz/controllers/multi_parameter_controller.py
```diff
@@ -65,19 +58,25 @@ def _update_histogram(parameter, hist_check_values, selected_ensembles):
...
-                ][parameter].data_df()
-                colors[parent.ensembles[ensemble_id]._name] = color["color"]
+            ensemble = load_ensemble(parent, ensemble_id)
+            if parameter in ensemble.parameters:
+                parameter_model = ensemble.parameters[parameter]
+                data[ensemble._name] = parameter_model.data_df()
+                colors[ensemble._name] = color["color"]
+
+                if parameter_model.priors and "prior" in hist_check_values:
+                    priors[ensemble._name] = (parameter_model.priors, color)
```
xjules, line 70: Shouldn't we use ids instead as we can have 2 ensembles with the same name?

ManInFez: yes wasnt sure if you had made that change in your pr

xjules: If you mean histogram then now, but can update it

ertviz/controllers/multi_parameter_controller.py
```diff
@@ -109,3 +108,29 @@ def _set_parameter_from_btn(_prev_click, _next_click, parameter_options, paramet
...
+        Output(parent.uuid("hist-check"), "options"),
+        [
+            Input(parent.uuid("parameter-selector"), "value"),
+        ],
+        [
+            State(parent.uuid("hist-check"), "options"),
+            State(parent.uuid("ensemble-selection-store"), "data"),
+        ],
+    )
+    def _set_parameter_from_btn(parameter, plotting_options, selected_ensembles):
```
xjules, line 122: Not sure, if this's necessary. Or if can you just make it disabled instead?

ertviz/controllers/multi_parameter_controller.py
xjules, line 70: Hm, do you think that prior can change over different ensembles for the same parameter?

xjules (approved):
Tested and approved :) Bra job!

PR #95, opened 2021-01-13
Add checkbox for showing only responses with obs

### LINKED ISSUE #80: Visualize in the dropdown which responses having observations
We should visualize or make the user in some way aware of which keys containing observations 

Should be done in the same way as in the legacy plotter with highlighting the background

autt:
Resolves #80 

Instead of highlighting the responses with observations as suggested in the issue, a checkbox is added such that only the responses with observations is shown when checked.

### DIFF
--- a/ertviz/controllers/multi_response_controller.py
+++ b/ertviz/controllers/multi_response_controller.py
@@ -88,13 +88,23 @@ def _create_response_plot(response, plot_type, selected_realizations, color):
     return ensemble_plot
 
 
+def _valid_response_option(response_filters, response):
+    if "obs" in response_filters:
+        return response.observations
+    else:
+        return True
+
+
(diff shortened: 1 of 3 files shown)


oysteoh:
Should this functionality also be included in the observation analyzer page? Maybe with the checkbox by default being selected? I’m not 100% sure of the functionality of this page, but i don't find it any useful when selecting parameters without observation - maybe they should just be all gone from the dropdown without any checkbox? 🤔

autt:
Yes, this functionality should also be included in the observation analyzer page. If there is no need to see the responses without observations, I guess they can be removed and that we don't need any checkbox :smile:

oysteoh (approved):
seems to work ! nice job 👍 👍  🚀

PR #97, opened 2021-01-14
Replace datetime.fromisoformat()

autt:
datetime.fromisoformat() is new in python 3.7 and does not run in
python<3.7.


oysteoh (approved):
seems to have fixed the problems for me at least! 🚀

Issue #98, opened 2021-01-14
Exceptions thrown for some responses in the observation analyzer page


oysteoh:
For some of the responses an exception is thrown when using a database from the HM tutorial. The same responses works perfectly fine in the other views.

I have the database-file if needed for investigation. Stacktrace below 
```
Traceback (most recent call last):
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/dash/dash.py", line 1059, in dispatch
    response.set_data(func(*args, outputs_list=outputs_list))
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/dash/dash.py", line 994, in add_context
    output_value = func(*args, **kwargs)  # %% callback invoked %%
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/ertviz/controllers/observation_response_controller.py", line 121, in _update_graph
    response_plots = [
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/ertviz/controllers/observation_response_controller.py", line 122, in <listcomp>
    _generate_plot(ensemble_id, data["color"])
  File "/Users/OOH/code/virtualenvs/ert/lib/python3.8/site-packages/ertviz/controllers/observation_response_controller.py", line 118, in _generate_plot
    plot = _create_misfits_plot(ensemble.responses[response], None, color)
[...]

PR #99, opened 2021-01-15
Convert datetime64 to float64 before adding to df

autt:
#98 
For some responses, the x_axis contains a datetime64. Pandas can not
convert datetime64 to float. Therefore, extract the date as a numpy
array and convert to float before adding to pandas dataframe.


xjules:
It solves the error, but I think we still need to update the labels on `x_axis`:
image

xjules (approved):
Good job! Have created an issue for the labels.

PR #104, opened 2021-01-19
Parallel coordinates

### LINKED ISSUE #85: Make use of parameter-parallel-coordinates from webviz subsurface
https://webviz-subsurface-example.azurewebsites.net/parameter-parallel-coordinates

xjules:
Resolves #85

- Ensemble gets a new method for exporting all parameters as a single dataframe
- new view "Parameter Comparison view"
- make use of `webvize-core-components` `wcc.Select`

### DIFF
--- a/ertviz/controllers/parameter_comparison_controller.py
+++ b/ertviz/controllers/parameter_comparison_controller.py
@@ -13,23 +13,30 @@ def parameter_comparison_controller(parent, app):
             "figure",
         ),
         [
-            Input(parent.uuid("parameter-selector-multi"), "value"),
+            Input(parent.uuid("parameter-selection-store"), "modified_timestamp"),
+            Input(parent.uuid("ensemble-selection-store"), "modified_timestamp"),
         ],
-        [State(parent.uuid("ensemble-selection-store"), "data")],
+        [
+            State(parent.uuid("parameter-selection-store"), "data"),
+            State(parent.uuid("ensemble-selection-store"), "data"),
+        ],
     )
-    def _update_parallel_coor(parameters, selected_ensembles):
-        if not selected_ensembles:
+    def _update_parallel_coor(
+        timestamp_param, timestamp_ensemble, selected_parameters, selected_ensembles
+    ):
+        if None in [selected_ensembles, timestamp_param, timestamp_ensemble]:
             raise PreventUpdate
+        selected_parameters = [] if not selected_parameters else selected_parameters
 
         data = {}
         colors = {}
-
         for idx, (ensemble_id, color) in enumerate(selected_ensembles.items()):
             ensemble = load_ensemble(parent, ensemble_id)
             ens_key = str(ensemble)
-            df = ensemble.parameters_df(parameters)
+            df = ensemble.parameters_df(selected_parameters)
             df["ensemble_id"] = idx
             data[ens_key] = df.copy()
             colors[ens_key] = color["color"]
         parent.parallel_plot = ParallelCoordinates(data, colors)
+
         return parent.parallel_plot.repr
--- a/ertviz/controllers/parameter_comparison_controller.py
+++ b/ertviz/controllers/parameter_comparison_controller.py
@@ -2,7 +2,7 @@ from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output, State
 from ertviz.models import (
     load_ensemble,
-    ParallelCoordinates,
+    ParallelCoordinatesPlotModel,
 )
 
 
@@ -21,10 +21,8 @@ def parameter_comparison_controller(parent, app):
             State(parent.uuid("ensemble-selection-store"), "data"),
         ],
     )
-    def _update_parallel_coor(
-        timestamp_param, timestamp_ensemble, selected_parameters, selected_ensembles
-    ):
-        if None in [selected_ensembles, timestamp_param, timestamp_ensemble]:
+    def update_parallel_coor(_, __, selected_parameters, selected_ensembles):
+        if selected_ensembles is None:
             raise PreventUpdate
         selected_parameters = [] if not selected_parameters else selected_parameters
 
@@ -37,6 +35,6 @@ def parameter_comparison_controller(parent, app):
             df["ensemble_id"] = idx
             data[ens_key] = df.copy()
             colors[ens_key] = color["color"]
-        parent.parallel_plot = ParallelCoordinates(data, colors)
+        parent.parallel_plot = ParallelCoordinatesPlotModel(data, colors)
 
         return parent.parallel_plot.repr
--- a/ertviz/controllers/parameter_comparison_controller.py
+++ b/ertviz/controllers/parameter_comparison_controller.py
@@ -23,6 +23,7 @@ def parameter_comparison_controller(parent, app):
 
         data = {}
         colors = {}
+
         for idx, (ensemble_id, color) in enumerate(selected_ensembles.items()):
             ensemble = load_ensemble(parent, ensemble_id)
             ens_key = str(ensemble)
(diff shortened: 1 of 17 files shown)


xjules:
Screenshot:
image

ertviz/controllers/parameter_comparison_controller.py
```diff
@@ -0,0 +1,35 @@
...
+        colors = []
+        for idx, (ensemble_id, color) in enumerate(selected_ensembles.items()):
+            ensemble = load_ensemble(parent, ensemble_id)
+            ens_key = str(ensemble)
+            df = ensemble.parameters_df(parameters)
+            df["ensemble_id"] = idx
+            data[ens_key] = df.copy()
+            colors.append([idx / len(selected_ensembles), color["color"]])
+            colors.append([(idx + 1) / len(selected_ensembles), color["color"]])
+        parent.parallel_plot = ParalellCoordinates(data, colors)
```
oysteoh, line 34: ParalellCoordinates -> ParallelCoordinates

ertviz/controllers/parameter_comparison_controller.py
```diff
@@ -0,0 +1,35 @@
...
+from ertviz.models import (
+    load_ensemble,
+    ParalellCoordinates,
+)
+
+
+def parameter_comparison_controller(parent, app):
+    @app.callback(
+        Output(
+            {"id": parent.uuid("paralell-coor"), "type": parent.uuid("graph")},
```
oysteoh, line 12: paralell-coor -> parallel-coor

ertviz/plugins/_parameter_comparison.py
```diff
@@ -0,0 +1,45 @@
...
+                html.Div(
+                    id=self.uuid("ensemble-content"),
+                    children=ensemble_selector_view(parent=self),
+                ),
+                html.Div(
+                    id=self.uuid("paralell-coor-content"),
+                    children=[
+                        html.H5("Multi parameter selector:"),
+                        parameter_selector_view(parent=self),
+                        paralell_coordinates_view(parent=self),
```
oysteoh, line 36: paralell_coordinates_view -> parallel_coordinates_view

ertviz/plugins/_parameter_comparison.py
```diff
@@ -0,0 +1,45 @@
...
+    @property
+    def layout(self):
+        return html.Div(
+            [
+                html.Div(
+                    id=self.uuid("ensemble-content"),
+                    children=ensemble_selector_view(parent=self),
+                ),
+                html.Div(
+                    id=self.uuid("paralell-coor-content"),
```
oysteoh, line 32: paralell-coor-content -> parallel-coor-content

ManInFez:
This looks really nice! I have a few things I would change:
1. Add place holder text for the regex input box
2. Put "Search string" on same line as input box
3. Remove "ParCoor", not necessary and takes up alot of space
4. After pressing enter in the input box, selecting all visible would be cool :) 
5. Clicking on an option should default to append to already selected. Otherwise easy to loose the previously selected.  
6. Add "clear selection" for both plot and options widget

xjules:
> This looks really nice! I have a few things I would change:

Thanks! :)

> 1. Add place holder text for the regex input box

Well, it's no really a regex (only a sub string currently). Shall we do it as regex though?

> 2. Put "Search string" on same line as input box

It was in the first version, no idea why I removed it ...

> 3. Remove "ParCoor", not necessary and takes up alot of space

I've left it there as a place holder if some checkboxes arrive at some point, but can go away.

> 4. After pressing enter in the input box, selecting all visible would be cool :)

Might be cool. It's not really a default behavior for the user  though; we can try it and left it as hint / comment.

> 5. Clicking on an option should default to append to already selected. Otherwise easy to loose the previously selected.
> 6. Add "clear selection" for both plot and options widget

I was thinking about two areas, one editable (to select from) and the other containing selected, but loose this idea as parameters are anyway visible as axes labels. Do you think about this as a default selection behavior, ie. only adding and one `clear all` button?

oysteoh:
I think also we should make the search box case-insensitive? At least i like that - but I understand if there is other opinions as well - i just think it offers better usability 😄

oysteoh:
And then i agree to what @ManInFez  says. And yes - i think it would be very useful with regexp - especially when there is a lot of params it would be really nice and offer great functionality !

Great job btw! 🚀

ManInFez:
I do not think we need full regex, but at least having wild cards, like the dcc.Dropdown uses space as wildcards :)

xjules:
> > 1. Add place holder text for the regex input box
> 
> Well, it's no really a regex (only a sub string currently). Shall we do it as regex though?

Done search syntax `"S1 S2 S3"` will get all parameters of type `aslkS1adlskjS2_eiuS3`
> 
> > 1. Put "Search string" on same line as input box
> 
> It was in the first version, no idea why I removed it ...

Done
> 
> > 1. Remove "ParCoor", not necessary and takes up alot of space
> 
> I've left it there as a place holder if some checkboxes arrive at some point, but can go away.
> 

Done
> > 1. After pressing enter in the input box, selecting all visible would be cool :)
> 
> Might be cool. It's not really a default behavior for the user though; we can try it and left it as hint / comment.
> 

Done, although need to fix the behavior of what's being selected and what's remained. 
> > 1. Clicking on an option should default to append to already selected. Otherwise easy to loose the previously selected.
> > 2. Add "clear selection" for both plot and options widget
> 

Done

ertviz/controllers/parameter_selector_controller.py
```diff
@@ -0,0 +1,83 @@
...
+            Input(parent.uuid("parameter-deactivator"), "value"),
+        ],
+    )
+    def update_parameters_options(selected_ensembles, filter_search, selected):
+        if not selected_ensembles:
+            raise PreventUpdate
+        selected = [] if not selected else selected
+
+        ensemble_id, _ = selected_ensembles.popitem()
+        ensemble = load_ensemble(parent, ensemble_id)
```
autt, line 38: Are parameter options equal for all ensembles, or do we want to make sure that all selected ensembles are considered when finding valid parameter options? 🤔 See #111 #108 #109

xjules: Good comment! Will need to make it as an intersection of ensemble parameters.

ertviz/controllers/parameter_selector_controller.py
```diff
@@ -0,0 +1,115 @@
+import regex
+import dash
+from dash.exceptions import PreventUpdate
+from dash.dependencies import Input, Output, State
+from ertviz.models import (
+    load_ensemble,
+)
+
+old_submit = 0
```
ManInFez, line 9: this is not in use anymore

ertviz/controllers/parameter_selector_controller.py
```diff
@@ -0,0 +1,119 @@
+import regex
+import dash
+from dash.exceptions import PreventUpdate
+from dash.dependencies import Input, Output, State
+from ertviz.models import (
+    load_ensemble,
+)
+
+
+def filter_match(filter, key):
```
ManInFez, line 10: `_filter_match`

ertviz/controllers/parameter_selector_controller.py
```diff
@@ -0,0 +1,119 @@
...
+            Output(parent.uuid("parameter-deactivator"), "value"),
+        ],
+        [
+            Input(parent.uuid("parameter-selection-store"), "modified_timestamp"),
+        ],
+        [
+            State(parent.uuid("parameter-selection-store"), "data"),
+        ],
+    )
+    def update_parameter_selection(ts, shown_parameters):
```
ManInFez, line 116: `ts` -> `_`

ertviz/models/plot_model.py
```diff
@@ -265,6 +266,61 @@ def repr(self):
         return fig
 
 
+class ParallelCoordinates:
```
ManInFez, line 269: `ParallelCoordinatesPlotModel`

ManInFez:
I think this looks great. Just some minor comments about naming style, but all in all very nice :)

ManInFez (approved):
Remeber to rebase and squash all non essential commits :)

PR #111, opened 2021-01-27
Fix responses

### LINKED ISSUE #108: Not all responses are shown when selecting comparison viewer compared to observation analyzer
Eivind had a case where not all responses was available in the comparison viewer but they was in the observations analyzer. Needs to be investigated.

autt:
Resolves #108 and #109 (hopefully)

Show union of valid responses for all selected ensembles in response selector


oysteoh (approved):
Seems to me like have fixed the problems - at least now i see all the responses when selecting multiple ensembles. Very good! 🚀

PR #112, opened 2021-01-29
Plot observations on top of all realizations

autt:


oysteoh (approved):
Yup - indeed the observations now are on the top! Great work ! 🥇

PR #115, opened 2021-02-02
Fix statistics plot color bug

ManInFez:


autt (approved):
Looks like the colors are correct now :+1:

Issue #113, opened 2021-02-02
Update legends in plotting page


ManInFez:
The legends in the plotting page are taking up a lot of space. We could remedy this by either:
- add show/hide legend toggle
- make text in legends more compact
- both

PR #117, opened 2021-02-03
Update legends

autt:
Solves #113 

Testing out checkbox for legend description and short repr name for ensembles.


ManInFez:
I am wondering if we should have this a single checkbox which works for all plots?

autt:
@ManInFez That might be a good idea. I will test it :+1:

oysteoh (changes requested):
This looks as it works like intended - good work! 🚀  

Is it necessary having all the commits, or could those be squashed into one or a couple of commits?

autt:
> This looks as it works like intended - good work! rocket
> 
> Is it necessary having all the commits, or could those be squashed into one or a couple of commits?

Squashed some commits, can squash more if wanted :smile:

oysteoh (approved):

PR #123, opened 2021-02-10
Add support for reponses in selector component

### LINKED ISSUE #122: Add support for response keys in Selector
As selector component should be used for selecting responses too, we should provide an option to display response keys.

xjules:
Resolves #122 

This adds an option to create selector with responses:
```python
parameter_selector_view(parent=self, data_type="response"),
```


oysteoh (approved):
🚀

PR #126, opened 2021-02-10
Add filter to remove keys ending with 'H'

oysteoh:


ertviz/controllers/controller_functions.py
```diff
@@ -1,8 +1,13 @@
...
+            return False
+
+    if "historical" in response_filters:
+        if response.name.split(":")[0][-1] == "H":
+            return False
+
+    return True
 
 
 def response_options(response_filters, ensembles):
```
xjules, line 13: Hm, was wondering if this function should include also a custom filers then (as a parameter)? For instance the regex.

xjules: Also it could be as an option to select whether we want to have an union or intersection among ensembles.

xjules: Let's make this as a separate PR.

xjules (approved):
Good job! LGTM!

PR #132, opened 2021-03-04
Color observations black when inactive

oysteoh:


ertviz/controllers/multi_response_controller.py
```diff
@@ -49,19 +49,26 @@ def _get_observation_plots(observation_df, x_axis):
     data = observation_df["values"]
     stds = observation_df["std"]
     x_axis = observation_df["x_axis"]
+    active_mask = observation_df["active"]
+
+    style = assets.ERTSTYLE["response-plot"]["observation"]
+    color = [style["color"] if active else "rgb(0, 0, 0)" for active in active_mask]
```
xjules, line 55: shall we use other color (some lighter tone of red) than black?

tests/plots/test_controller.py
```diff
@@ -26,6 +26,7 @@ def test_observation_plot_representation():
             "values": [2.85325093, 7.20311703, 21.38648991, 31.51455593, 53.56766604],
             "std": [0.1, 1.1, 4.1, 9.1, 16.1],
             "x_axis": [0, 2, 4, 6, 8],
+            "active": [True, False, True, False, False],
```
xjules, line 29: Can you make it part of the `assert` test / section too, ie. testing text values as plot property?

ertviz/controllers/multi_response_controller.py
oysteoh: We can - i tried lighter red but i didn't get it to "stand out" as a deactivated one.... tried grey as well - but it is kind of difficult.... you have any suggestions?

xjules: Did some experiments: `rgb(255,127,0)` looks pretty decent.
image

oysteoh: agree - stand out at least! But - does it communicate `inactive` or `deactivated`?  🤔  From the look of it i cannot interpret if it is an active obs or not. With black / grey / faded / more transparent my brain immediately thinks something is turned off...

xjules (approved):
Bra job! :rocket:

PR #133, opened 2021-03-05
Add attribute text as hovertext for observations

oysteoh:


ertviz/models/observation.py
```diff
@@ -8,17 +8,24 @@ def __init__(self, observation_schema):
...
+
         if "data" in observation_schema:
             data = observation_schema["data"]
             self._x_axis = data["x_axis"]["data"]
             self._std = data["std"]["data"]
             self._values = data["values"]["data"]
 
+        if "attributes" in observation_schema:
+            for k, v in observation_schema["attributes"].items():
+                self._attributes += f"{k}: {v}<br>"
```
xjules, line 21: is this like continuous string which comes as parameter?

tests/plots/test_controller.py
```diff
@@ -26,6 +26,7 @@ def test_observation_plot_representation():
             "values": [2.85325093, 7.20311703, 21.38648991, 31.51455593, 53.56766604],
             "std": [0.1, 1.1, 4.1, 9.1, 16.1],
             "x_axis": [0, 2, 4, 6, 8],
+            "attributes": "Key1 Value1<br>Key2 Value2<br>",
```
xjules, line 29: Was just wondering if this is part of some test or it will fail if `attributes` are no present.

ertviz/controllers/multi_response_controller.py
```diff
@@ -49,19 +49,21 @@ def _get_observation_plots(observation_df, x_axis):
     data = observation_df["values"]
     stds = observation_df["std"]
     x_axis = observation_df["x_axis"]
+    attributes = observation_df["attributes"]
```
xjules, line 52: What if attributes are not present? Could we make like 
```python
if "attributes" in observation_df:
   ...
```

tests/plots/test_controller.py
oysteoh: added assertion and squashed commits 👍

xjules (approved):
LGTM and bra job! :)

PR #135, opened 2021-03-15
Add help tour_steps for correlation view plugin

### LINKED ISSUE #134: Add help guide for interacting with correlation view plugin
`Response Correlation Viewer` requires some level of interaction and therefore we should add a context help.

xjules:
Resolves #134

This exploits the `tour_steps` property of the plugin. 
- [x] make it jump across views when doing the tour


oysteoh:
highlight does not fit the scatter chart ( all others works fine ). Is this possible to fix?

oysteoh (approved):

PR #147, opened 2021-03-29
Add min-height attribute style for container that have issues

### LINKED ISSUE #146: Plots get disappearing in plugins
Plots do get disappear or covered other-views. It should be enough to set min-height for a container that include the plot.

xjules:
Resolves #146 
- [x] Plot viewer
- [x] Response correlation


oysteoh (approved):
LGTM!

PR #153, opened 2021-04-23
New dataloader

dotfloat:


webviz_ert/data_loader/__init__.py
```diff
@@ -79,10 +130,81 @@ def get_parameter_data_url(
...
+
+class DataLoader:
+    _instances: MutableMapping[ServerIdentifier, "DataLoader"] = {}
+
+    baseurl: str
+    token: str
+    _graphql_cache: MutableMapping[str, MutableMapping[dict, Any]]
+
+    def __new__(cls, baseurl: str, token: Optional[str] = None) -> "DataLoader":
+        if (baseurl, token) in cls._instances:
```
xjules, line 146: Not sure that I understand this branch here.

dotfloat: I want exactly one `DataLoader` object for each `baseurl` and `token` pair. `DataLoader` should be able to do caching before this PR is merged.

Using `__new__` instead of `__init__` because it still allows us to use `DataLoader(..., ...)` to initialise the object, _but_ `__new__` is called before the object is instantiated, so we can do instantiation manually and return the same object if the arguments are the same. `__init__`, on the other hand, first allocated the objects before calling `__init__` to instantiate it.

webviz_ert/data_loader/__init__.py
```diff
@@ -1,88 +1,242 @@
-from typing import Mapping, Any, Union, Mapping, Optional, List
 import os
+import json
+from typing import Mapping, Any, Union, Mapping, Optional, List, MutableMapping, Tuple
```
oysteoh, line 3: Mapping not used

oysteoh (approved):
👍

PR #156, opened 2021-06-01
Replace priors endpoint with graphql query

xjules:
Part of the process of removing prior endpoint.
Links to https://github.com/equinor/ert-storage/issues/119


oysteoh (approved):
LGTM! 🚀

PR #158, opened 2021-06-07
Fix parameters ert3

### LINKED ISSUE #157: Plot single record containing multiple parameters
**Background**
Since ert3 allows to define parameters as a group, eg. 
```yaml
-
  name: coefficients
  type: stochastic
  distribution:
    type: gaussian
    input:
      mean: 0
      std: 1
  variables: ["a", "b", "c"]
```
they are stored as `record_class=ds.RecordClass.parameter` and  `record_type=ds.RecordType.f64_matrix,`, where record column labels represent variable names `["a", "b", "c"]` and index column defines different realizations. Currently, `webviz-ert` doesn't visualize such type of parameters.

**Approach**
Create a dedicated `PlotModel` class which would be able to handle such parameters. An initial approach could be just to create (automatically) multiple histogram views (separate graphs), ie. for `a`,`b` and `c` for a parameter that is called `coefficients`.

Link to the accompanied discussion https://github.com/equinor/ert/issues/1720 that spawned this issue.

xjules:
Resolves #157 

It incorporates the following changes in order that this PR works:
- [x] rename metadata to userdata
- [ ] spawns independent plots for the compound parameter model

### DIFF
--- a/tests/data/snake_oil_data.py
+++ b/tests/data/snake_oil_data.py
@@ -67,7 +67,7 @@ ensembles_response = {
                 "id": 1,
                 "timeCreated": "2020-04-29T09:36:26",
                 "size": 1,
-                "Metadata": '{"name": "default"}',
+                "userdata": '{"name": "default"}',
             }
         }
     },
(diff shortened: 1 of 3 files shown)


xjules:
Closing this as we will handle it in ert.

PR #160, opened 2021-06-11
Pass token to requests in dataloader

### LINKED ISSUE #155: Use token-based auth for ERT Storage
Now that tokens are in, we should add token-based auth to webviz. Probably depends on https://github.com/equinor/ert/issues/1700

ManInFez:
resolves #155

### DIFF
--- a/webviz_ert/data_loader/__init__.py
+++ b/webviz_ert/data_loader/__init__.py
@@ -14,15 +14,17 @@ logger = logging.getLogger()
 def get_info(project_id: str = None) -> Mapping[str, str]:
     from ert_shared.storage.connection import get_info
 
-    return get_info(project_id)
+    info = get_info(project_id)
+    info["auth"] = info["auth"][1]
+    return info
 
 
 # these are needed to mock for testing
-def _requests_get(*args: Union[str, bytes], **kwargs: Any) -> requests.models.Response:
+def _requests_get(*args: Any, **kwargs: Any) -> requests.models.Response:
     return requests.get(*args, **kwargs)
 
 
-def _requests_post(*args: Union[str, bytes], **kwargs: Any) -> requests.models.Response:
+def _requests_post(*args: Any, **kwargs: Any) -> requests.models.Response:
     return requests.post(*args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


webviz_ert/data_loader/__init__.py
```diff
@@ -14,15 +14,17 @@
...
 
 
 # these are needed to mock for testing
-def _requests_get(*args: Union[str, bytes], **kwargs: Any) -> requests.models.Response:
+def _requests_get(*args: Any, **kwargs: Any) -> requests.models.Response:
     return requests.get(*args, **kwargs)
 
 
-def _requests_post(*args: Union[str, bytes], **kwargs: Any) -> requests.models.Response:
+def _requests_post(*args: Any, **kwargs: Any) -> requests.models.Response:
```
xjules, line 27: `Union[str, bytes,dict]`?

webviz_ert/data_loader/__init__.py
```diff
@@ -150,6 +153,11 @@ def _query(self, query: str, **kwargs: Any) -> dict:
     def _get(
         self, url: str, headers: dict = None, params: dict = None
     ) -> requests.Response:
+        if headers is None:
+            headers = {"Token": self.token}
+        else:
+            headers.update({"Token": self.token})
+
         resp = _requests_get(f"{self.baseurl}/{url}", headers=headers, params=params)
```
dotfloat, line 161:
```suggestion
        if headers is None:
            headers = {}

        resp = _requests_get(f"{self.baseurl}/{url}", headers={**headers, "Token": self.token}, params=params)
```

1. Avoids code duplication
2. Avoids mutating input data

webviz_ert/data_loader/__init__.py
ManInFez: fails with `error: Argument 1 to "post" has incompatible type "*Tuple[Union[str, bytes, Dict[Any, Any]], ...]"; expected "bool"  [arg-type]`

xjules: ok, then `Any` fits the bill.

dotfloat (approved):

PR #161, opened 2021-06-15
Fix observation style

### LINKED ISSUE #145: Active and inactive observations in various response plots is wrong
**Describe the bug**
The active / inactive observations as indicated by color-coding in various plots throughout webviz-ert is wrong, or at least incompatible with the updatelog for the project. The indicator might be correct when only one response is selected, but when more than one response is plotted confusion appear. See screenshots showing the same case with one response and four responses. This goes both for the Plot Viewer, and the Response Correlation Viewer. To me it seems like the mask is extracted from one response and applied to all plots, but not sure. 

A related issue (I think), is that when selecting several responses with different number of observations, some plots are empty. In my case when plotting WWCT:* four responses with 3 observations each are plotted, but the one with 4 is not displayed, and I'm guessing it is because the active/inactive mask has a different dimensionality that the observation vector.   

**Expected behavior**
That color coding represent the truth no matter how many responses are plotted together :)

**Screenshots**
Capture4
Capture3

 - RGS RHEL7
 - Chrome

xjules:
Resolves #145 
When changing observation style for inactive and active observations, we need to copy the style first.

### DIFF
--- a/webviz_ert/controllers/multi_response_controller.py
+++ b/webviz_ert/controllers/multi_response_controller.py
@@ -74,7 +74,7 @@ def _get_observation_plots(observation_df: pd.DataFrame) -> PlotModel:
     attributes = observation_df["attributes"]
     active_mask = observation_df["active"]
 
-    style = assets.ERTSTYLE["response-plot"]["observation"]
+    style = deepcopy(assets.ERTSTYLE["response-plot"]["observation"])
     color = [style["color"] if active else "rgb(0, 0, 0)" for active in active_mask]
     style["marker"]["color"] = color
 


webviz_ert/controllers/multi_response_controller.py
```diff
@@ -74,7 +74,7 @@ def _get_observation_plots(observation_df: pd.DataFrame) -> PlotModel:
     attributes = observation_df["attributes"]
     active_mask = observation_df["active"]
 
-    style = assets.ERTSTYLE["response-plot"]["observation"]
+    style = assets.ERTSTYLE["response-plot"]["observation"].copy()
```
ManInFez, line 77: should this perhaps be a `deepcopy`?

xjules: I thought about it, will do so :+1:

ManInFez (approved):

PR #164, opened 2021-07-09
Sort response dataframes by x_axis

ManInFez:
resolves https://github.com/equinor/ert/issues/1811


ManInFez:
This addresses the problem of timeseries data coming in unsorted on the plotting side. It could be argued that it should be addressed on the side which is posting data. However, the plotting should ignore the NaNs in the dataframe and plot a connected line when there are missing values.

sondreso (approved):

PR #188, opened 2021-11-02
Enable support for ensembles with no data inside

### LINKED ISSUE #187: Strange behavior for ES-MDA example case
**Describe the bug**
When displaying plots for the case "default" the page gets refreshed and the front page view is displayed.

**To Reproduce**

Steps to reproduce the behaviour:
1. Run an es-mda poly-example run with `--ensemble-new-storage` 
2. Start webviz-ert: `ert vis poly.ert`
3. Select `Plot Viewer` or `Response Correlation Viewer` 
4. In the Ensemble selector select the `default` ensemble  
5. Select one parameter or response to display

**Expected behaviour**
The correct ensemble data should be displayed

DanSava:
Resolves #187

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -1,5 +1,6 @@
 from typing import List, Optional, Dict, Mapping, Tuple, Any
 
+import pandas as pd
 from dash.development.base_component import Component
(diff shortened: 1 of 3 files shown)


DanSava:
Jenkins test this please

oysteoh (approved):

Issue #191, opened 2021-11-23
Unable to plot coefficients in ERT3 polynomial example


markusdregi:
**Describe the bug**
When trying to visualise the coefficients from the ERT3 polynomial example I get a stack trace in the terminal and no data visualised.

**To Reproduce**
1. Install ert and webviz-ert, both `main`
2. Launch ert-storage via `ert3 service start storage`
3. Create example with `ert3 init --example polynomial`
4. Enter workspace with `cd polynomial`
5. Run experiment with `ert3 run evaluate`
6. Launch visualisation with `ert vis`
7. Select `Plot Viewer` in sidebar
8. Select the `evaluation` experiment
9. Select the `coefficients` under `Parameters`
10. Observe the crash

**Expected behavior**
Plots displaying the coefficients of the different realisations.

**Stack trace**
```
Traceback (most recent call last):
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/dash/dash.py", line 1336, in dispatch
    response.set_data(func(*args, outputs_list=outputs_list))
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/dash/_callback.py", line 151, in add_context
    output_value = func(*func_args, **func_kwargs)  # %% callback invoked %%
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/webviz_ert/controllers/multi_parameter_controller.py", line 73, in update_histogram
    data[key] = parameter_model.data_df()
  File "/Users/mfane/Development/demo/demo-py37/env/lib/python3.9/site-packages/webviz_ert/models/parameter_model.py", line 32, in data_df
[...]

PR #192, opened 2021-11-23
Account for ert and ert3 discrepancy when uploading parameter data to storage

DanSava:
**Issue**
Resolves: #191 

It seems parameters are handled differently for ert3 and ert at the moment:

* **Ert3**: uploads a record for each realisation index.
* **Ert**: uploads only one global record for each parameter. 

**Approach**

For a given parameter (i.e. coefficients) try to retrieve a global record (assuming the user has run ERT) if this fails for the same parameter try to retrieve a record for a realization index in the list of active realization of the run example.

### DIFF
--- a/webviz_ert/data_loader/__init__.py
+++ b/webviz_ert/data_loader/__init__.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Mapping, Optional, List, MutableMapping, Tuple
+from typing import Any, Mapping, Optional, List, MutableMapping, Tuple, Dict
 from collections import defaultdict
 from pprint import pformat
 import requests
@@ -113,7 +113,7 @@ class DataLoader:
     _instances: MutableMapping[ServerIdentifier, "DataLoader"] = {}
 
     baseurl: str
-    token: str
+    token: Optional[str]
     _graphql_cache: MutableMapping[str, MutableMapping[dict, Any]]
 
     def __new__(cls, baseurl: str, token: Optional[str] = None) -> "DataLoader":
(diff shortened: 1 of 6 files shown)


markusdregi:
I did some testing of this PR and I can confirm that the visualiser no longer crashes 👍🏻

First, a comment on the behaviour of the visualiser: `coefficients` consists of three different variables, namely `a`, `b` and `c`. I believe that earlier I've been able to plot these independently and this no longer seems possible...

Second, a comment on the approach: I'm not entirely convinced that this should be resolved in `webviz-ert` given the apparent reason for the problem. If `ert-storage` serves the records differently dependent on whether they where uploaded in one batch or one by one I would perhaps consider that an `ert-storage` bug?

webviz_ert/data_loader/__init__.py
```diff
@@ -191,32 +191,62 @@ def get_ensemble_userdata(self, ensemble_id: str) -> dict:
...
+                resp = self._get(
+                    url=f"ensembles/{ensemble_id}/records/{parameter_name}",
+                    headers={"accept": "application/x-parquet"},
+                )
+            stream = io.BytesIO(resp.content)
+            df = pd.read_parquet(stream)
+            return df
+        except DataLoaderException as e:
+            logger.error(e)
+            return self.get_ensemble_record_data(
```
mortalisk, line 226: Is this necessary because the above only fetches ensemble wide data, and this call fetches realization by realization?

DanSava: Yes, that's right

frode-aarstad (approved):
LGTM

webviz_ert/models/ensemble_model.py
```diff
@@ -98,7 +101,18 @@ def parameters(
         self,
     ) -> Optional[Mapping[str, ParametersModel]]:
         if not self._parameters:
-            parameter_names = self._data_loader.get_ensemble_parameters(self._id)
+            parameter_names = []
+            # TODO Remove except code path after PR https://github.com/equinor/ert-storage/pull/180 is merged
```
markusdregi, line 105: I think TODO's are more likely to be picked up in issues, then in the code ;)

markusdregi:
I'm pressuming this is not the end of this effort? Would be nice to see the issues pointing laying out the remaining work before we merge this :)

webviz_ert/models/ensemble_model.py
DanSava: Generally, that is right but the point of this TODO was to remind me that there is still something to be solved and removed before having a final version of the code and merging it.... like a TODO is meant to be right :thinking: ?

markusdregi: Ah, so this is not ready for merging yet then? :)

DanSava: It is now that the TODO was dealt with

DanSava:
> I'm pressuming this is not the end of this effort

Not really sure what "this effort" is that you are hinting at. 

This PR should enable webviz ert to properly display grouped ert3 parameters and plot them also get responses as ensemble wide records.

DanSava:
Jenkins test this please

webviz_ert/data_loader/__init__.py
```diff
@@ -113,7 +113,7 @@ class DataLoader:
     _instances: MutableMapping[ServerIdentifier, "DataLoader"] = {}
 
     baseurl: str
-    token: str
+    token: Optional[str]
```
markusdregi, line 116: Could you explain why this is optional now?

webviz_ert/data_loader/__init__.py
```diff
@@ -195,45 +195,62 @@ def get_ensemble_userdata(self, ensemble_id: str) -> dict:
...
     ) -> pd.DataFrame:
-        resp = self._get(
-            url=f"ensembles/{ensemble_id}/records/{parameter_name}",
-            headers={"accept": "application/x-parquet"},
-        )
-        stream = io.BytesIO(resp.content)
-        df = pd.read_parquet(stream)
-        return df
+        try:
+            if "::" in parameter_name:
```
markusdregi, line 212: This is what I meant by future work. It might be that I'm misunderstanding something, but is this not somewhat of a hack to avoid having to make ERT2 upload parameters as it should?

markusdregi (changes requested):
I was expecting some tests verifying the current `::` logic...?

markusdregi:
I did some testing and it seems like the original problems are fixed 👍🏻 I tried running the entire polynomial case with ert3 using `run_demo` and it crashes when I'm trying to visualise the `doe` experiment with

```
raise DataLoaderException(
webviz_ert.data_loader.DataLoaderException: Error fetching data from http://127.0.0.1:51822/ensembles/12d84e2f-7ccc-4e95-bd1a-328c0658b962/records/coefficients/labels
                The request return with status code: 404
                b'{"detail":{"ensemble_id":"12d84e2f-7ccc-4e95-bd1a-328c0658b962","name":"coefficients","error":"Record not found"}}'
```

I guess we should either fix it here or create a new issue for it. I'm fine with either :)

webviz_ert/data_loader/__init__.py
DanSava: Not really sure why pylint was not complaining before about this, it just started, I guess it was updated recently. 

The Dataloader `__new__` definition looked like this for quite some time 

```
def __new__(cls, baseurl: str, token: Optional[str] = None) -> "DataLoader":
```

So token is expected to he `Optional[str]`

markusdregi: Thanks :)

webviz_ert/data_loader/__init__.py
DanSava: Sure we can make ert2 upload parameters like ert3 does it, but I thought the idea was that the underlying Data structure is the same for both ert3 and ert2 it should not matter how they are uploaded, ert2 or ert3 style given that both ways are valid and supported by the storage API.

DanSava:
> I did some testing and it seems like the original problems are fixed 👍🏻 I tried running the entire polynomial case with ert3 using `run_demo` and it crashes when I'm trying to visualise the `doe` experiment with
> 
> ```
> raise DataLoaderException(
> webviz_ert.data_loader.DataLoaderException: Error fetching data from http://127.0.0.1:51822/ensembles/12d84e2f-7ccc-4e95-bd1a-328c0658b962/records/coefficients/labels
>                 The request return with status code: 404
>                 b'{"detail":{"ensemble_id":"12d84e2f-7ccc-4e95-bd1a-328c0658b962","name":"coefficients","error":"Record not found"}}'
> ```
> 
> I guess we should either fix it here or create a new issue for it. I'm fine with either :)

I think a new issue is required, I will create it. I think this never worked (the visualisation) it cannot find the `coefficient` record because it is not part of the doe ensemble due to it being added to storage as part of the "general" ensemble. 

The coefficients are added to storage before the experiment is run with `ert3 record load designed_coefficients experiments/doe/coefficients_record.json`

DanSava:
https://github.com/equinor/webviz-ert/issues/203 was added to track the issue reported :arrow_up:

markusdregi (approved):
Thanks for the update 👍🏻

Issue #195, opened 2021-12-07
X-axis on parameters in the plot view is not ordered correctly


oysteoh:

DanSava:
I don't think there is an issue with the parameters x-axis:
There might be a small issue with the y-axis but it is not critical in my opinion.

**Side by side comparison QT plotter with Webviz-ERTside by side comparison QT plotter with Webviz-ERT**:
![Screenshot from 2021-12-08 13-49-53](https://user-images.githubusercontent.com/4508053/145203986-4b41c3ae-a353-4c0a-bfe3-70a45ffb48a2.png)

The shape of the parameter histogram seems to be correct so I am not sure what the problem would be with the x-axis of the parameters for the webviz-ert case. 

But there seems to be an issue with the x-axis for the responses in the plot view

PR #201, opened 2021-12-08
Remove record data index sorting not really needed 

DanSava:
**Issue**
Index sorting is also prone to issues when sorting ints that have been converted to strings

Resolves: #195

### DIFF
--- a/webviz_ert/data_loader/__init__.py
+++ b/webviz_ert/data_loader/__init__.py
@@ -219,7 +219,7 @@ class DataLoader:
                     params={"realization_index": rel_idx},
                 )
                 stream = io.BytesIO(resp.content)
-                df = pd.read_parquet(stream).transpose().sort_index()
+                df = pd.read_parquet(stream).transpose()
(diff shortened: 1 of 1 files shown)


frode-aarstad:
I am not really able to reproduce what this was supposed to fix. I remember @DanSava showing a webviz plot that had this problem when discussing perfomance. Could you please tell me how you made that dataset.

oysteoh:
> I am not really able to reproduce what this was supposed to fix. I remember @DanSava showing a webviz plot that had this problem when discussing perfomance. Could you please tell me how you made that dataset.

At least I can reproduce and verify it is actually working. I edited the poly-case to run `ensemble_experiment` with 20 iterations instead of 10. In the `Plot Viewer` select the ensemble and then the `response` `poly_res@0`. Without Dan's fix i will get the x-axis like this `0, 1, 10, 11 ... 19, 2 .. 9` - with Dan's fix it is in the correct order.

frode-aarstad (approved):

PR #215, opened 2022-02-09
Cache ensemble model application level not page level

### LINKED ISSUE #213: Improve ensemble caching
**Is your feature request related to a problem? Please describe.**
Currently, ensemble models are cached per page (i.e. Plot Viewer, Response Correlation Viewer, etc). This means when a page is opened by a user for the first time the ensemble models are created if that page was not opened before, even though the same ensemble model is common for all pages. 

**Describe the solution you'd like**
Consider caching at the application level

DanSava:
**Issue**
Resolves #213

**Approach**
Cache ensemble model at the application level

### DIFF
--- a/webviz_ert/controllers/observation_response_controller.py
+++ b/webviz_ert/controllers/observation_response_controller.py
@@ -113,8 +113,8 @@ def observation_response_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
                     selection=None
                 )  # What about selections?
                 if summary_df is not None:
-                    data_dict[parent.ensembles[ensemble_id]._name] = summary_df
-                colors[parent.ensembles[ensemble_id]._name] = color["color"]
+                    data_dict[ensemble._name] = summary_df
+                colors[ensemble._name] = color["color"]
             if data_dict:
                 plot = MultiHistogramPlotModel(
                     data_dict,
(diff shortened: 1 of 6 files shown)


webviz_ert/controllers/observation_response_controller.py
```diff
@@ -113,8 +113,8 @@ def update_graph(
                     selection=None
                 )  # What about selections?
                 if summary_df is not None:
-                    data_dict[parent.ensembles[ensemble_id]._name] = summary_df
-                colors[parent.ensembles[ensemble_id]._name] = color["color"]
+                    data_dict[ensemble._name] = summary_df
```
oysteoh, line 116: Should we do a refactoring / expose name as a property in EnsembleModel class avoid the use of private members?

oysteoh:
Looking good - tested on a fairly big case and now it is actually usable! Really nice improvement to the performance ! Had one minor comment regarding exposure/consuming `private` _name member of the EnsembleModel - except that i'm ready to approve :rocket:

webviz_ert/controllers/observation_response_controller.py
DanSava: Good idea, I avoided doing it as part of this PR given that it was not really related to what I was trying to do. But I don't mind doing it as part of this PR if you are also OK with it.

DanSava: Done

oysteoh (approved):

PR #220, opened 2022-02-16
Remove front page and load plot viewer directly

oysteoh:


DanSava:
If we remove the mention in `webviz-config.yml`, do we still need `webviz_ert/plugins/_ensemble_overview.py` and if not I guess the tests also need to be updated.

oysteoh:
> If we remove the mention in `webviz-config.yml`, do we still need `webviz_ert/plugins/_ensemble_overview.py` and if not I guess the tests also need to be updated.

guess not - cannot see any reason for it .... I will remove 👍

frode-aarstad (approved):
LGTM

PR #222, opened 2022-02-21
Improve usability of the ensemble selector

### LINKED ISSUE #217: Improve usability of the case view
The current case view is very hard to navigate. Also, the discoverability of how to select multiple cases is low (based on user feedback). I suggest we replace it with a simple select list for now.

DanSava:
**Issue**
Resolves #217

**Approach**
Fix min and max zoom levels
Add tooltip information for selection of multiple ensembles

If not enough last resort use the same hack as parameter or result selection controllers


oysteoh:
I think it is at least much better with this approach limiting the zoom and also adding some tooltip to indicate functionality. That said, not sure if it still looks good, but this has been discussed before - it is hard to come up with something which works nice here! But - lets discuss tomorrow! Also - the minimize seems a bit weird - ending up like a "transparent frame" down right bottom with me - but now not able to zoom to select anything. Minimize should maybe just collapse it into a button or something ( if we choose to continue with this approach as you propose ).

webviz_ert/controllers/ensemble_selector_controller.py
```diff
@@ -7,155 +7,80 @@
 from webviz_ert.models import load_ensemble
 import webviz_ert.assets as assets
```
oysteoh, line 8: not used anymore

oysteoh:
Just a general comment while quickly looking through the changes, a lot of unused imports in the controllers - not related to your changes but might as well clean that up in a separate commit ?

And - cytoscape i think can be removed from requirements?

DanSava:
> Just a general comment while quickly looking through the changes, a lot of unused imports in the controllers - not related to your changes but might as well clean that up in a separate commit?

There are a lot of things that need to be clean up in this repo, but yes maybe a separate commit with at least cleaning imports is not a bad idea. 

> And - cytoscape i think can be removed from requirements?

Yes, I think it can be removed. I was not aware it was a requirement :open_mouth:

oysteoh:
I think this looks good for now! We can discuss further if we should move it / change layout, but functionality is what we was looking for and i think now this is a much better solution ! :rocket:

oysteoh (approved):

Issue #209, opened 2022-01-05
Redrawing responses takes time


DanSava:
**Issue**
Every time a new response is selected or deselected all previously drawn responses are resized and redrawn. 

**Describe the solution you'd like**
Investigate if it is possible to reduce drawing time for responses, also if possible just resize previous drawn responses and avoid redrawing plots when possible.

DanSava:
Did some testing for pure draw times of dash and plotly using the following example:

``` python
import time

import dash
from dash import html, dcc
from dash.dependencies import Output, Input, State
import pandas as pd
import plotly.graph_objs as go
import random
import numpy as np
import plotly.express as px

# create app
app = dash.Dash(
    __name__,
    external_stylesheets=[
        "https://cdnjs.cloudflare.com/ajax/libs/bulma/0.7.2/css/bulma.min.css"
    ],
)

fig = go.Figure()
reals = 1000
nu_vals = 100


app.layout = html.Div(
    className="section",
    id="main_layout",
    children=[
        html.Button(
            id="enter-button1",
            className="button is-large is-outlined",
            children=["Test Scattergl"],
        ),
        dcc.Link("ccc", id="id_link", href="/test"),
        html.Label(id="lbl_2", children=["test"]),
        html.P(""),
        html.Button(
            id="enter-button2",
            className="button is-large is-outlined",
            children=["Test plotly express line"],
        ),
        dcc.Link("ccc", id="id_link2", href="/test"),
        html.Label(id="lbl_1", children=["test"]),
    ],
)


@app.callback(
    [Output("id_link", "children"), Output("lbl_2", "children")],
    [Input("enter-button1", "n_clicks")],
    [State("id_link", "children")],
)
def button_1(n_clicks, _):
    if n_clicks is not None:
        fig.data = []
        start_time = time.time()
        for i in range(reals):
            fig.add_trace(
                go.Scattergl(
                    x=list(range(nu_vals)),
                    y=[random.randint(0, nu_vals) for _ in range(nu_vals)],
                )
            )
        fig_gen_time = time.time() - start_time
        return [dcc.Graph(figure=fig)], [
            f"Reals: {reals} Values: {nu_vals} --- Total time = {fig_gen_time}"
        ]
    return [dcc.Graph(figure=fig)], [""]


@app.callback(
    [Output("id_link2", "children"), Output("lbl_1", "children")],
    [Input("enter-button2", "n_clicks")],
    [State("id_link2", "children")],
)
[...]

PR #230, opened 2022-03-08
Small performance improvement for the graph grid when adding and removing plots

DanSava:
**Issue**
Related to  #209

**Approach**
Stop using an extra dash store to keep track of displayed plots.


oysteoh (approved):
I think this actually makes it much better! Running on RGS with the Drogoncase makes a big differnence i would say! :rocket:

Issue #231, opened 2022-03-08
Sort selection lists


markusdregi:
All selection lists in webviz-ert should be sorted alphabetically. That includes the case list, the response list and the parameter list.

PR #233, opened 2022-03-09
Sort the responses and parameter views

frode-aarstad:
**Issue**
Resolves equinor/webviz-ert#231

### DIFF
--- a/webviz_ert/controllers/observation_response_controller.py
+++ b/webviz_ert/controllers/observation_response_controller.py
@@ -127,7 +127,8 @@ def observation_response_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
 
         def _generate_plot(ensemble_id: str, color: str) -> ResponsePlotModel:
             ensemble = load_ensemble(parent, ensemble_id)
-            plot = _create_misfits_plot(ensemble.responses[response], [], color)
+            resp = ensemble.responses[str(response)]
+            plot = _create_misfits_plot(resp, [], color)
             return plot
 
         response_plots = [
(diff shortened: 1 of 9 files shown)


webviz_ert/controllers/observation_response_controller.py
```diff
@@ -127,7 +127,8 @@ def update_graph(
 
         def _generate_plot(ensemble_id: str, color: str) -> ResponsePlotModel:
             ensemble = load_ensemble(parent, ensemble_id)
-            plot = _create_misfits_plot(ensemble.responses[response], [], color)
+            resp = ensemble.responses[str(response)]
```
DanSava, line 130: Is this refactoring really needed? Also, why is the string conversion required ? `str(response)`?

webviz_ert/models/ensemble_model.py
```diff
@@ -2,8 +2,8 @@
 import pandas as pd
 from typing import Mapping, List, Dict, Union, Any, Optional
 from webviz_ert.data_loader import get_data_loader, DataLoaderException
-
 from webviz_ert.models import Response, PriorModel, ParametersModel
+from collections import OrderedDict
```
DanSava, line 6: Is this import used?

DanSava (approved):

webviz_ert/controllers/observation_response_controller.py
DanSava: Still, something is fishy here that you had to explicitly cast to string... Any way this function is a bit messy I think it should be refactored in the future such that it does not take `response` from the outer scope but it should expect a response name as part of the function arguments.

PR #237, opened 2022-03-31
Remove search from some dropdowns

### LINKED ISSUE #235: Response / parameter search confusing
There is a `search` bar available which works as expected, but users ( test-users never seen this before ) attempts to write in the drop-down selection box underneath where the selected responses / parameters goes. We should disable the box from input ( and why is it indicated as a dropdown? It seems to just get filled with selected keys? 
Skjermbilde 2022-03-29 kl  12 03 41
)

frode-aarstad:
**Issue**
Resolves #235 

Removed search functionality from dropdowns for ensemble and parameters (same as responses). 
Removed "Select ..." text in the box which is misleading.


oysteoh (approved):
LGTM! 🚀

PR #239, opened 2022-04-05
Catch exceptions

### LINKED ISSUE #229: Clear cache / recover from failure in selected data
User testing with datasets new / unknown to us sometimes leads to errors in webviz-ert - typically a key is selected which causing an exception. The user then gets redirected immediately to the front-page. Going back to the page in question the key causing the failure is already selected and naturally causing the same exception again and the user is thrown back to the front-page without the possibility of deselecting the key. 

Solving this issue i guess is 2 part
- Handling what is actually causing the exception in webviz-ert  
- Always recover from an exception and / or notify the user of the event with the possibility to at least remove or do something with the faulty data instead of being forced to close and then launch `ert vis ..... .ert` again.

frode-aarstad:
**Issue**
Resolves #229

### DIFF
--- a/webviz_ert/data_loader/__init__.py
+++ b/webviz_ert/data_loader/__init__.py
@@ -150,9 +150,6 @@ class DataLoader:
             raise RuntimeError(
                 f"ERT Storage query returned with '{resp.status_code}':\n{pformat(doc)}"
             )
-
-        # self._graphql_cache[query][kwargs] = doc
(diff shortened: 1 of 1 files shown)


oysteoh (approved):
Handling exceptions improve the user experience a lot. This should be merged. How we are presenting the errors should be discussed with our users / webviz team. Created a separate issue for this and added it to our backlog https://github.com/equinor/webviz-ert/issues/241

PR #242, opened 2022-05-23
Add button to trigger ensemble data refresh

DanSava:
**Issue**
Closes https://github.com/equinor/ert/issues/3355

**Approach**
Add a button to trigger the new storage endpoint `/updates/facade`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -68,8 +68,8 @@ from webviz_ert.data_loader import GET_ENSEMBLE
 
 
 def _requests_post(url, **kwargs):
-    payload = kwargs["json"]
-    if payload["query"] == GET_ENSEMBLE:
+    payload = kwargs["json"] if "json" in kwargs else None
+    if payload is not None and payload["query"] == GET_ENSEMBLE:
         variables = payload["variables"]
         ensemble_id = variables["id"]
         url = f"http://127.0.0.1:5000/ensembles/{ensemble_id}"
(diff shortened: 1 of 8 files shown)


oysteoh:
Clicking on the refresh button makes all the selections cleared and reverted back to initial state, but the charts remains open . A bit confusing.... :thinking:

oysteoh:
Looking better now i think! But - adding more ensembles appending a full set of the selected keys.

Screenshot from 2022-05-24 08-46-16


Screenshot from 2022-05-24 08-46-31

DanSava:
Jenkins test this

tests/conftest.py
```diff
@@ -82,7 +82,7 @@ def _requests_post(url, **kwargs):
 def select_first(dash_duo, selector):
     parameter_selector_input = dash_duo.find_element(selector)
     options = parameter_selector_input.text.split("\n")
-    dash_duo.click_at_coord_fractions(parameter_selector_input, 0.1, 0.1)
+    dash_duo.click_at_coord_fractions(parameter_selector_input, 0.1, 0.05)
```
DanSava, line 85: Had to change this because when using this function to select the first element from the parameter selection list it would actually select the second one

tests/data/snake_oil_data.py
```diff
@@ -108,6 +109,7 @@ def content(self):
     "http://127.0.0.1:5000/ensembles/2/parameters": [
         {"name": "test_parameter_1", "labels": []},
         {"name": "test_parameter_11", "labels": []},
+        {"name": "BPR_138_PERSISTENCE", "labels": []},
```
DanSava, line 112: Had to add an extra parameter to the second ensemble because I needed two ensembles with the same parameter

oysteoh (approved):
LGTM! :rocket:  Just remember squash your commits :+1:

PR #255, opened 2022-05-25
Fix overlapping parameter plot headers

### LINKED ISSUE #245: Fix overlapping titles
Long titles on the keys will overlap in the plot viewer when adding more charts to the view. We should make it look nice.

frode-aarstad:
**Issue**
Resolves #245 

- Added a ... when name is too large.
- Added a tooltip with the complete parameter name
- Refactored the "options" to be on its own row below header


DanSava (approved):
Looks good :+1:

PR #257, opened 2022-05-26
Add new dropdown control for parameters and responses

### LINKED ISSUE #254: You can select multiple responses but only one is plotted. It should restrict to only one selected. todo

DanSava:
**Issue**
Was supposed to address  #254 before the issue was understood a bit more. Now using a single response selector does not seem to be the way to go.


DanSava:
Jenkins test this

DanSava:
No longer needed by #254

PR #259, opened 2022-05-30
Show ensemble color for each selected ensemble

### LINKED ISSUE #246: Specify which ensemble belongs to which color
In some of the plots, adding more ensembles to a chart could be confusing. They have different colors, but the legend have the same information regarding the datapoints across the various ensembles - which makes it impossible to know which ensemble have which color.

DanSava:
**Issue**
Closes #246

## Pre review checklist


DanSava:
Jenkins test this please

oysteoh:
Jenkins: Test this please!

oysteoh (commented):
Looks very good - think this has solved the issue perfectly fine! One "question" ... I'm not sure if it make sense - but - should the text on hover of the different lines be changed from `Realization-{realization}-{ensemble_name}`  to something like `Realization: {realization}, Ensemble: {ensemble_name}` ?  I find this `default`  "default" naming annoying just popping up. It is the default name yes, but just using the name without context could mean so much more.... 

I.e - current situation
`Realization-0-default_0`

suggested
`Realization: 0, Ensemble: default_0`


And maybe the same on the legends which now is `R_0-default_0` to `R: 0, E: default_0` ?

DanSava:
> I.e - current situation `Realization-0-default_0`
> 
> suggested `Realization: 0, Ensemble: default_0`

I think this is a brilliant suggestion :1st_place_medal: 

> 
> And maybe the same on the legends which now is `R_0-default_0` to `R: 0, E: default_0` ?

I am not sure about  `R: #, E: default_#` to be honest I don't really like `R_#-default_#` either. Now that the ensemble's colour is clear from the ensemble dropdown selection, I would like to go back to `Realization-# ` for the legend and only keep having the additional information regarding the ensemble on the line hover.

oysteoh:
> I am not sure about R: #, E: default_# to be honest I don't really like R_#-default_# either. Now that the ensemble's colour is clear from the ensemble dropdown selection, I would like to go back to Realization-# for the legend and only keep having the additional information regarding the ensemble on the line hover.

Agree :+1:

tests/views/test_ensemble_selector.py
```diff
@@ -83,3 +83,33 @@ def test_ensemble_refresh(
         dash_duo, "#" + plugin.uuid("ensemble-multi-selector")
     )
     assert refreshed_cases == expected_cases
+
+
+def test_ensemble_color(mock_data, dash_duo, tmpdir):
```
oysteoh, line 88: doesn't matter i guess  - but `tmpdir` is not used as far as i can see?

oysteoh (approved):
LGTM! :rocket:

tests/views/test_ensemble_selector.py
DanSava: Good catch! it is there from a failed attempt to run the test in a temp folder and copy the `ert-style.css` file there. But that did not seem to work for some reason and I abandoned that route. 

I will remove it before merge

Issue #228, opened 2022-03-04
Add  button or any other way of collapsing the new ensemble selection list


DanSava:
**Issue**
The new ensemble selection list cannot be collapsed, it would be consistent to be collapsable like the parameter selection list.

Might be also solved by: #227

PR #260, opened 2022-05-30
WIP: Redesign search fields

### LINKED ISSUE #250: Search fields 3 columns side-by-side, one toggle visibility button
Reorganize the search boxes and have all 3 side-by-side with the possibility of toggling all on/off

valentin-krasontovitsch:
**Issue**
Closes #250 and #228 

**Approach**
We rearrange the ensemble, response and parameter areas into one row, and have one button that toggles visibility for all three, across all plugins / views / whatever you call the different sites available through the menu on the left

## Pre review checklist


ertomatic:

valentin-krasontovitsch:
besides the strange exception logsin observation analyzer I think this is ready for review... gonna wait and see if we can figure that one out though

valentin-krasontovitsch:
The tests are failing - I'm not sure why. I am confident though (tried locally at least) that they will pass once the axes labels PR is in, with a slight refactoring of the tests, and this work gets rebased on top of that.

gonna switch from work in progress to ready for review.

**NB reviewers**: Feel free to review commit by commit, i tried to package the commits nicely, and have desriptive commit messages (which will of course be squashed laterz).

valentin-krasontovitsch:
yesss, tests are passing now.

tests/data/spe1_st/tests/test_webviz_ert.py
```diff
@@ -37,7 +37,7 @@ def test_webviz_parameter_comparison(get_ensemble_id, dash_duo):
 
     ensemble_name = select_first(dash_duo, "#" + plugin.uuid("ensemble-multi-selector"))
     _verify_keys_in_menu(
-        dash_duo, plugin, parameter_keys, "parameter-selector-multi-params"
+        dash_duo, plugin, parameter_keys, "parameter-selector-multi-param"
```
oysteoh, line 40: why the renaming from params -> param?

webviz_ert/views/ensemble_selector_view.py
```diff
@@ -29,11 +29,13 @@ def ensemble_selector_list(parent: WebvizPluginABC) -> List[Component]:
                     dbc.Col(
                         html.Button(
                             id=parent.uuid(f"parameter-selector-button"),
-                            children=("Toggle selector visibilities"),
+                            children=("Hide Selectors"),
```
oysteoh, line 32: Why change name ? I guess the button actually `toggles` and not only `hide` the selectors?

tests/data/spe1_st/tests/test_webviz_ert.py
valentin-krasontovitsch: because i think a jenkins test fails otherwise

so this is a bit of a problem - because I'm not sure if I was just to eager, or if I did good, but I remember seeing an inconsistency with respect to using `param` vs. `params` - and in particular where the string was given as `suffix`, it was `param`, so I believe I changed it to `param` in some places. but i didn't see the extent of it (there are many places where it's still params).

so i've encountered the exact same problem in another PR and "fixed" it there in the same spot, too... perhaps this needs a proper clean up, in a separate issue

valentin-krasontovitsch: oh yeah and I don't understand why that jenkins test suddenly started failing

the change seems to be due to 3bc13f629ea0f890bf7796679aacdef539dffafb though, where the suffix "params" was replaces with `DataType.PARAMETERS` which has `value`: `param`

but how that change got merged, and now jenkins is failing, is a riddle to me

maybe it didn't fail on this PR either, and I'm just making up that memory - maybe i just happened to grep for params and thought that I missed it, or something!? I don't know, I'm sorry :disappointed:

webviz_ert/views/ensemble_selector_view.py
valentin-krasontovitsch: `Hide Selectors` is only the initial name - logic was added to switch the name between `hide` and `show`, and the reason for removing the word visibilities is a visual one - the button becomes smaller

tests/data/spe1_st/tests/test_webviz_ert.py
valentin-krasontovitsch: hmm looking at the changes I see that I changed it in many many spots, and not just in the line you're pointing out :sweat_smile: 

In that case, you can partly disregard the bla bla up there, and my answer is actually: to make things more consistent. maybe this shouldn't have been part of this PR though.

If you wish, I can revert the renames (where that's possible without breaking things after a rebase), just let me know.

valentin-krasontovitsch: yupp after rebase all these renames are gone! gonna resolve this : )

oysteoh:
we are getting there ! but - still some nitpicking 😉  The ones marked in red not longer have any function and should be removed i think? 
Skjermbilde 2022-07-04 kl  19 44 59


And - now seeing it - i start to think - would it make sense using icons here and have expand / collapse / refresh together on the top  and maybe have some kind of "framing" of the three selectors indicating they are together? 
Skjermbilde 2022-07-04 kl  19 49 28

valentin-krasontovitsch:
oh yeah they should not be there :thinking: weird! gonna check that out, good catch.

i can try with some icons, and see what it looks like - perhaps in a different branch, so we can compare

valentin-krasontovitsch:
do we have an equinor icon library? would that be part of ... that equinor design project, i forget its name?

valentin-krasontovitsch:
looking into the icon aspect, I've discovered the buttonbar on the upper right of our pages - exploring the code surrounding that one, it seems to be non-trivial to add a custom button there (need to modify / piggyback a webviz core component).

trying to just display icons from the equinor design system also appears difficult - we are writing python code, those are available as react components, and it is not clear to me how i can inject a react component into a webviz plugin.

it seems like webviz itself uses the equinor design system and in particular its icons, based on a quick search of the code base, but again it is unclear to me how I can hijack / piggyback on their imports.

I'm gonna try one more time to see if I can find some "default" or "standard" icons that can be used from dash

oysteoh:
> so one can use [icons in dash](https://dash-bootstrap-components.opensource.faculty.ai/docs/icons/), but these are based on bootstrap, and equinor icons are based on material design, so this would constitute a design inconsitency. further more, one needs to tell dash to load a custom css scripts, and I could sadly not figure out where we call `dash.Dash` - it is passed as an argument to our plugins, but I don't see where it's called, and hence where I could try to inject an external stylesheet. finally a complete vanilla approach without anything doesn't seem possible - and I don't want to paste the literal SVGs for the icons needed, that just doesn't seem right.
> 
> i've asked in the interal communication channels for support on webviz, but I think for now I will have to abandon the icon idea, sorry

No problems - buttons with text will do it for now :+1:

tests/views/test_plot_view.py
```diff
@@ -7,6 +8,8 @@
     select_response,
     wait_a_bit,
 )
+from webviz_ert.plugins import *
```
oysteoh, line 11: usually not a fan of * imports , any reasons for not being spesific?

oysteoh:
I think this looks nice now - one minor comment i want answered. And then rebase and squash and i'll approve :+1:

tests/views/test_plot_view.py
valentin-krasontovitsch: hmm I've never thought much about the advantages or disadvantages of the `*` import in python, so it might be due to simple lack of critical thinking / knowledge. Basically I just though I want to check all plugins, so I'm gonna import all of them.

i can change it and list the names, if you'd like / think it's better?

valentin-krasontovitsch: fixed! the second line in particular seems to have been redundant.

valentin-krasontovitsch:
comment addressed, now i just gotta make the tests pass

valentin-krasontovitsch:
... and rebase + resolve conflicts :sweat_smile:

oysteoh (approved):
:rocket:

PR #277, opened 2022-06-13
Add hint to use config file during start up

valentin-krasontovitsch:
**Issue**
This came up while trying to start webviz ert for the first time.

Following the instructions from the readme to the dot, starting webviz-ert doesn't necessarily work - I had to supply a config file.

**Approach**

After discussing a bit in the team, we've concurred that adding this hint should be harmless.

## Pre review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,4 +35,8 @@ ert ensemble_smoother --target-case smoother poly.ert
 
 # After simulation has finished, start webviz-ert from the same location with
 ert vis
+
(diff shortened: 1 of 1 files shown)


valentin-krasontovitsch:
Also, I still don't understand the difference between ert-storage, dark storage, and classic ert (legacy storage?) - so if my comment can be made more precise, please tell me how, and i will update it

DanSava (approved):
:+1:

PR #288, opened 2022-06-20
Remove type from correlation_view

### LINKED ISSUE #273: Refactor correlation_view
The correlation_view Graph element has the following id:

```python

{
    "id": id_view,
    "type": parent.uuid("graph"),
}

```

Did a very quick check and `type` seems to be redundant and not really used for anything, check if this is actually the case by removing it.

The resulting graph element in `webviz_ert/views/correlation_view.py` should look something like this:

```python 
  dcc.Graph(
      id=id_view,
      className="ert-view-cell",
      config={"responsive": True},
  )

```

hnformentin:
**Issue**
Resolves #273

**Approach**
Solution indicated by @Dan.

### DIFF
--- a/webviz_ert/views/correlation_view.py
+++ b/webviz_ert/views/correlation_view.py
@@ -5,16 +5,13 @@ from dash import html
 from dash import dcc
 
 
-def correlation_view(parent: WebvizPluginABC, id_view: str) -> Component:
+def correlation_view(id_view: str) -> Component:
     return html.Div(
-        id=id_view,
+        id=f"container_{id_view}",
         className="ert-view-container",
         children=[
             dcc.Graph(
-                id={
-                    "id": id_view,
-                    "type": parent.uuid("graph"),
-                },
+                id=id_view,
                 className="ert-view-cell",
                 config={"responsive": True},
             )
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -27,17 +27,11 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
     @app.callback(
         [
             Output(
-                {
-                    "id": parent.uuid("response-correlation"),
-                    "type": parent.uuid("graph"),
-                },
+                parent.uuid("response-correlation"),
                 "figure",
             ),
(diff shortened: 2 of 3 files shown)


webviz_ert/views/correlation_view.py
```diff
@@ -11,10 +11,7 @@ def correlation_view(parent: WebvizPluginABC, id_view: str) -> Component:
         className="ert-view-container",
         children=[
             dcc.Graph(
-                id={
-                    "id": id_view,
-                    "type": parent.uuid("graph"),
-                },
+                id=id_view,
```
DanSava, line 14: Took a quick look at your PR and it seems the problem in the test happens because two components have the same id, one of them is the graph  `dcc.Graph(id=id_view, ...)` and the other one is the `Div` that contains the graph see lines 9 and 10 in this file  `webviz_ert/views/correlation_view.py`

```python
html.Div(
  id=id_view
```

Dash does not allow two components with the same ids hence the issues you have been seeing in the tests. 

possible fix replace

```
html.Div(
  id=id_view
```
with

```
html.Div(
  id=f"container_{id_view}"
```

Also the function `correlation_view` does not require the `parent` argument anymore.

DanSava (commented):
You should rebase on main otherwise the jenkins tests will not pass

webviz_ert/views/correlation_view.py
hnformentin: Thanks! The solution seems good. What would you think if we have `view_container` instead of `container` only?
Removing `parent` from `correlation_view` makes the test to fail locally while with it they pass.
Here both fails, so I will investigate...

hnformentin: I did not find a reason why the test `test_axes_labels` was failing. When I tried to run again, it passed. I am not sure if this is a flakiness that should be investigate.

hnformentin:
After (and if) approved, I will rebase in one commit before merging.

hnformentin:
Jenkins is failing in `test_axes_labels`, I will create an issue to solve the flakiness in this test and hold this PR until the issue is solved.

oysteoh (approved):
seems to be stable and work as intended now. Not able to reproduce the failing test.

PR #290, opened 2022-06-21
Separate plots in plot viewer

### LINKED ISSUE #268: Plot viewer should be split in two
One for "responses" and one for "parameters", or at least it is a bit "messy" having distributions, summary data etc in a big lump. We would rather have one summary data, and only be able to select one response, now it needs two clicks to change the view.

We decided during the planning to make an horizontal separation between the plots.

valentin-krasontovitsch:
**Issue**
Closes #268 

**Approach**
We arrange the response plots in one row, and the parameter plots in another row.
To this end, we parametrize the parameter / response duality in the plot view controller instead of handling both at once, and call two instances of the controller, one for responses and one for parameters, respectively.

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/plot_view_controller.py
+++ b/webviz_ert/controllers/plot_view_controller.py
@@ -11,79 +11,77 @@ import webviz_ert.assets as assets
 
 from webviz_ert.views import response_view, parameter_view
 from webviz_ert.plugins import WebvizErtPluginABC
+from webviz_ert.models import DataType
 
 
-def _new_child(parent: WebvizErtPluginABC, plot: Dict) -> Component:
-    if plot["type"] == "response":
-        p = response_view(parent=parent, index=plot["name"])
-    elif plot["type"] == "parameter":
-        p = parameter_view(parent=parent, index=plot["name"])
-    else:
-        raise ValueError(f"Plots of undefined type {plot['type']}")
+def _new_child(parent: WebvizErtPluginABC, plot: str, data_type: DataType) -> Component:
+    if data_type == DataType.RESPONSE:
+        p = response_view(parent=parent, index=plot)
+    if data_type == DataType.PARAMETER:
+        p = parameter_view(parent=parent, index=plot)
 
     return dbc.Col(
-        html.Div(id=parent.uuid(plot["name"]), children=p),
+        html.Div(id=parent.uuid(plot), children=p),
         xl=12,
         lg=6,
         style=assets.ERTSTYLE["dbc-column"],
-        key=plot["name"],
+        key=plot,
     )
 
 
-def plot_view_controller(parent: WebvizErtPluginABC, app: dash.Dash) -> None:
-    webviz_ert.controllers.multi_response_controller(parent, app)
-    webviz_ert.controllers.multi_parameter_controller(parent, app)
+def plot_view_controller(
+    parent: WebvizErtPluginABC, app: dash.Dash, data_type: DataType
+) -> None:
+    if data_type == DataType.PARAMETER:
+        webviz_ert.controllers.multi_parameter_controller(parent, app)
+    elif data_type == DataType.RESPONSE:
+        webviz_ert.controllers.multi_response_controller(parent, app)
+    else:
+        raise Exception(f"Unexpected data type `{data_type}`")
 
     @app.callback(
-        Output(parent.uuid("plot-selection-store"), "data"),
+        Output(parent.uuid(f"plot-selection-store-{data_type}"), "data"),
         [
-            Input(parent.uuid("parameter-selection-store-param"), "modified_timestamp"),
-            Input(parent.uuid("parameter-selection-store-resp"), "modified_timestamp"),
+            Input(
+                parent.uuid(f"parameter-selection-store-{data_type}"),
+                "modified_timestamp",
+            ),
         ],
         [
-            State(parent.uuid("parameter-selection-store-param"), "data"),
-            State(parent.uuid("parameter-selection-store-resp"), "data"),
-            State(parent.uuid("plot-selection-store"), "data"),
+            State(parent.uuid(f"parameter-selection-store-{data_type}"), "data"),
+            State(parent.uuid(f"plot-selection-store-{data_type}"), "data"),
         ],
     )
     def update_plot_selection(
         _: Any,
-        __: Any,
-        parameters: Optional[List[str]],
-        responses: Optional[List[str]],
-        current_selection: Optional[List[Mapping[str, str]]],
-    ) -> List[Mapping[str, str]]:
-        parameters = [] if not parameters else parameters
-        responses = [] if not responses else responses
-        current_selection = [] if not current_selection else current_selection
-        for plot in current_selection.copy():
-            if plot["name"] not in parameters and plot["name"] not in responses:
-                current_selection.remove(plot)
-        for param in parameters:
-            new_plot = {"name": param, "type": "parameter"}
-            if new_plot not in current_selection:
-                current_selection.append(new_plot)
-        for response in responses:
-            new_plot = {"name": response, "type": "response"}
-            if new_plot not in current_selection:
-                current_selection.append(new_plot)
+        selection: Optional[List[str]],
+        current_plots: Optional[List[str]],
+    ) -> List[str]:
+        selection = [] if not selection else selection
+        current_plots = [] if not current_plots else current_plots
+        for plot in current_plots.copy():
+            if plot not in selection:
+                current_plots.remove(plot)
+        for selected in selection:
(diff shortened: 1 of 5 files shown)


valentin-krasontovitsch:
I have not added any tests - and the tests all passed (at least locally) with the changes I introduced. Not sure if it's necessary to test that the plots are indeed separated visually, I feel like maybe not.

Regarding the changes in the controller and testing thereof - it seems that the controllers in general are not very extensively tested at the moment.

perhaps that's a problem with the design, with those callbacks and whatnot, might be difficult to hook into that, I don't know - if that is the case, they could be refactored to export the logic into more testable functions, and then tests could be added.

Maybe that's outside of the scope of this issue / PR though. But I could create an issue, if anybody else seconds that we should have moar tests for the controllers.

webviz_ert/controllers/plot_view_controller.py
```diff
@@ -30,60 +38,59 @@ def _new_child(parent: WebvizErtPluginABC, plot: Dict) -> Component:
...
-                current_selection.append(new_plot)
-
-        return current_selection
+        selection = [] if not selection else selection
+        current_plots = [] if not current_plots else current_plots
+        for plot in current_plots.copy():
+            if plot["name"] not in selection:
+                current_plots.remove(plot)
+        for selected in selection:
+            new_plot = {"name": selected, "type": _suffix_to_plot_type(suffix)}
```
DanSava, line 74: Because of the split, I don't think keeping track of the plot type is required, and the current_plot can be a simple list of strings and not a list of dicts

webviz_ert/controllers/plot_view_controller.py
```diff
@@ -30,60 +38,59 @@ def _new_child(parent: WebvizErtPluginABC, plot: Dict) -> Component:
...
+            if new_plot not in current_plots:
+                current_plots.append(new_plot)
+
+        return current_plots
```
DanSava, line 78: Probably, I would have just sorted the current plots by type :sunglasses: but I think it is nice to have the actual separation between responses and parameters.

webviz_ert/controllers/plot_view_controller.py
```diff
@@ -13,6 +13,14 @@
...
 
+def _suffix_to_plot_type(suffix: str) -> str:
+    if suffix == "resp":
+        return "response"
+    if suffix == "param":
+        return "parameter"
+    raise Exception(f"Unexpected suffix `{suffix}`")
+
+
 def _new_child(parent: WebvizErtPluginABC, plot: Dict) -> Component:
```
DanSava, line 24: I think the `suffix` or hopefully the `DataType` is available when you call this function. If you added  it to the list of arguments the function definition would look something like

```
def _new_child(parent: WebvizErtPluginABC, plot_name: str, data_type: DataType) -> Component:
```

webviz_ert/controllers/plot_view_controller.py
```diff
@@ -13,6 +13,14 @@
 from webviz_ert.plugins._webviz_ert import WebvizErtPluginABC
 
 
+def _suffix_to_plot_type(suffix: str) -> str:
+    if suffix == "resp":
+        return "response"
+    if suffix == "param":
+        return "parameter"
+    raise Exception(f"Unexpected suffix `{suffix}`")
```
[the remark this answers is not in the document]
valentin-krasontovitsch: which part? the exception? or the function? it makes the code easier to read, I feel, to extract this in a bit of own logic, we can test it (hurr hurr), and easily extend it and use it elsewhere... just feels tidier to me, but maybe it's just my feelings and it's premature optimization, or something like that. i can leave that block (without the exception) in the function that calls this one, too, just feel like it's a nice little stand alone thing that can be pulled out (for great success).
what downsides am i not seeing, because i am often blind for them once i'm convinced that it should be like this or that 😅?

webviz_ert/controllers/plot_view_controller.py
```diff
@@ -30,60 +38,59 @@ def _new_child(parent: WebvizErtPluginABC, plot: Dict) -> Component:
     )
 
 
-def plot_view_controller(parent: WebvizErtPluginABC, app: dash.Dash) -> None:
-    webviz_ert.controllers.multi_response_controller(parent, app)
-    webviz_ert.controllers.multi_parameter_controller(parent, app)
+def plot_view_controller(
+    parent: WebvizErtPluginABC, app: dash.Dash, suffix: str
```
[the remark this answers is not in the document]
valentin-krasontovitsch: ah yes, very well, sure! I like that : )

can I move the function `_suffix_to_plot_type` to the file where DataType is defined, and adjust it accordingly? that is, if i get to keep it ; )

webviz_ert/controllers/plot_view_controller.py
valentin-krasontovitsch: ah is that what you meant above as well? i'll look into it, thanks for the hint! just assumed that the info was used here or there...

valentin-krasontovitsch: awesome you were right, thanks for catching that!
first part (removing the type) done in d0bec85d4ad9e2
gonna look at `DataType` next

webviz_ert/controllers/plot_view_controller.py
valentin-krasontovitsch: cool thank you ^^

webviz_ert/controllers/plot_view_controller.py
valentin-krasontovitsch: and done completely with 59b97b9

webviz_ert/controllers/plot_view_controller.py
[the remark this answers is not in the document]
valentin-krasontovitsch: done in 59b97b9!

webviz_ert/controllers/plot_view_controller.py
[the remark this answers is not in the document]
valentin-krasontovitsch: removed in 59b97b9

valentin-krasontovitsch:
Jenkins: Please test this!

webviz_ert/controllers/plot_view_controller.py
```diff
@@ -11,79 +11,77 @@
...
+            if selected not in current_plots:
+                current_plots.append(selected)
+
+        return current_plots
 
     @app.callback(
-        Output(parent.uuid("plotting-content"), "children"),
-        Input(parent.uuid("plot-selection-store"), "data"),
-        State(parent.uuid("plotting-content"), "children"),
+        Output(parent.uuid(f"plotting-content-{data_type.value}"), "children"),
```
DanSava, line 72: I don't think you need {data_type.value} I think {data_type} should be enough here.

webviz_ert/views/parameter_view.py
```diff
@@ -9,7 +9,7 @@
 import webviz_ert.assets as assets
 
 
-def parameter_view(parent: WebvizPluginABC, index: int = 0) -> List[Component]:
+def parameter_view(parent: WebvizPluginABC, index: str = "") -> List[Component]:
```
DanSava, line 12: Why did you change this index from int to str ?

webviz_ert/views/response_view.py
```diff
@@ -7,7 +7,7 @@
 import dash_bootstrap_components as dbc
 
 
-def response_view(parent: WebvizPluginABC, index: int = 0) -> List[Component]:
+def response_view(parent: WebvizPluginABC, index: str = "") -> List[Component]:
```
DanSava, line 10: Why did you change this index from int to str ?

DanSava (approved):
Looks good :+1:, had one more small comment and some questions but overall looks good. Don't forget to squash the commits before merging

webviz_ert/controllers/plot_view_controller.py
valentin-krasontovitsch: ya awesome indeed! i must have misunderstood the docs... thanks!
fixed.

webviz_ert/views/parameter_view.py
valentin-krasontovitsch: because it seems like we are using a string when we call this function, and `mypy` complaints if we leave it as `int` in the function signature :thinking: 
i tried to figure out how and what, but then I just quickly tested if this works, and it makes mypy happy, and nobody else complained either, and the site still worked fine. but i did not try to figure out if this actually needs to be an int...

valentin-krasontovitsch: i'm gonna go ahead and merge, as no exceptions were thrown, all tests passed, and no functionality seemed removed based on manual visual testing due to this particular change.

PR #294, opened 2022-06-28
Revert experimental flag in ResponseCorrelation ParameterComparison

### LINKED ISSUE #281: Revert experimental state
**Issue**

In `webviz_ert/assets/webviz-config.yml` for `ParameterComparison` and `ResponseCorrelation` set `experimental` to `False`

We have been a bit trigger-happy with the use of the newly implemented ability to set a page as `experimental`  and this is what happened for  `ParameterComparison` and `ResponseCorrelation` but once a page is set as experimental it will not be displayed by default and maybe we need a bit more time to discuss which of the pages should actually be experimental.

hnformentin:
**Issue**
Resolves #281

### DIFF
--- a/webviz_ert/assets/webviz-config.yml
+++ b/webviz_ert/assets/webviz-config.yml
@@ -13,9 +13,9 @@ pages:
  - title: Parameter Comparison Viewer
    content:
     - ParameterComparison:
-   experimental: True
+   experimental: False
 
  - title: Response Correlation Viewer
    content:
     - ResponseCorrelation:
-   experimental: True
+   experimental: False


webviz_ert/assets/webviz-config.yml
```diff
@@ -13,9 +13,7 @@ pages:
  - title: Parameter Comparison Viewer
    content:
     - ParameterComparison:
-   experimental: True
```
DanSava, line 16: This works :+1:  but I was thinking of just replacing `experimental: True` with `experimental: False` so that we know there is functionality for setting a page experimental and it is optional and can be activated by setting the flag to `True`

hnformentin: I see, in this case, for consistency, I would add the flag `experimental: False` to all of them. Is it ok?

DanSava: Actually, no the point is that this flag is optional if you add it for all the pages then if somebody new reads the config file in a few months will think that they need to set the flag for a new page as well which is not true.

hnformentin: I see the point. I think it makes sense since we want to add the flag `beta` on the experimental views, if I understand correctly.

DanSava (approved):

PR #305, opened 2022-07-04
Show realization number in misfit plots

### LINKED ISSUE #289: Add realization number to Ert vis Misfit/Observations plot
Hovering over the Mistfit/Observation plots in ert vis gives access to the actual misfit numbers and box plot stats but not the realization number. There are situations where getting this information is valuable. 

This should ideally apply to both the individual numbers in the "scatter" plot in the box plots, and to lower plot in the histogram version.

valentin-krasontovitsch:
**Issue**
Closes #289 

**Approach**
This concerns two different plots in the misfits plugin aka Observation Analyzer - the univariate view (bunch of boxplots), and the summary histograms.

For the bosplots, we use `customdata` to supply the realization numbers and insert it into the hover text by overriding the `hovertemplate`.

For the summary, #92 already did most of the work - we only needed to transpose the misfits dataframe in question.

Please test this with data, in particular also with several ensembles chosen (as the example I used for manual testing only had one ensemble).

## Pre review checklist

### DIFF
--- a/tests/plots/test_controller.py
+++ b/tests/plots/test_controller.py
@@ -177,13 +177,19 @@ def test_parallel_coordinates_representation():
 def test_univariate_misfits_boxplot_representation():
     data = np.random.rand(200).reshape(-1, 20)
     missfits_df = pd.DataFrame(data=data, index=range(10), columns=range(20))
-    plots = _get_univariate_misfits_boxplots(missfits_df.copy(), color="rgb(255,0,0)")
+    ensemble_name = "test-ensemble"
+    plots = _get_univariate_misfits_boxplots(
+        missfits_df.copy(), ensemble_name=ensemble_name, color="rgb(255,0,0)"
+    )
     assert len(plots) == 20
     for id_plot, plot in enumerate(plots):
         np.testing.assert_equal(0.3, plot.repr.jitter)
         np.testing.assert_equal("all", plot.repr.boxpoints)
         x_pos = missfits_df.columns[id_plot]
-        name = f"{x_pos}"
+        if isinstance(x_pos, int):
+            name = f"Value {x_pos}"
+        else:
+            name = f"{x_pos} - {ensemble_name}"
         assert name == plot.repr.name
 
 
--- a/webviz_ert/controllers/observation_response_controller.py
+++ b/webviz_ert/controllers/observation_response_controller.py
@@ -19,28 +19,33 @@ from webviz_ert import assets
 
 
 def _get_univariate_misfits_boxplots(
-    misfits_df: Optional[pd.DataFrame], color: str
+    misfits_df: Optional[pd.DataFrame], ensemble_name: str, color: str
 ) -> List[BoxPlotModel]:
     if misfits_df is None:
         return []
     x_axis = misfits_df.columns
-    misfits_data = list()
+    realization_names = [f"Realization {name}" for name in misfits_df.index.values]
+    misfit_plots = list()
     for misfits in misfits_df:
         plot = BoxPlotModel(
             y_axis=misfits_df[misfits].abs().values,
-            name=f"{misfits}",
+            name=misfits,
+            ensemble_name=ensemble_name,
             color=color,
+            customdata=realization_names,
+            hovertemplate="(%{x}, %{y}) - %{customdata}",
         )
-        misfits_data.append(plot)
-    return misfits_data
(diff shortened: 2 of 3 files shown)


valentin-krasontovitsch:
Jenkins: please test this.

oysteoh:
Looks good - one minor comment / question. Is it possible also stating which ensemble this is to avoid confusion if the colors are quite similar / help those with color-blindness? 

Screenshot from 2022-07-05 08-08-04

valentin-krasontovitsch:
ya - good points, I'm sure, let me look into that!

oysteoh:
Getting there 👍 👍  I thought i mentioned it, but seems i forgot. Hover the actual boxes shows no information except from data ( min, max, mean etc. ) Should those as well have some info regarding realisation / ensemble? I think it would be nice - having multiple ensembles on top of eachother makes it a bit hard to know what you are looking at.

valentin-krasontovitsch:
OK two things - 

- i forgot to look into adding ensemble name to the legend
- i hoped to get away without modifying the boxplots, as I don't know (yet) how to modify the text there : )

gonna look into those now!

maybe i can use this to ask a quick question - can you give me some example commands to run in a local test data example to produce data for several ensembles? so i can test this properly? ^^

valentin-krasontovitsch:
little clarification - do I understand correctly that the box itself cannot show information about the realizations, as the box itself (OBS: using layman's terms, not technical terms) is a statistic over the values of the different realizations, or a summary? 

the `boxpoints` display the discrete values, and display which realization gave the corresponding value.

the boxplot can say which ensemble it belongs to though.

valentin-krasontovitsch:
okay force pushed (after rebase to latest main) two more commits, one a small variable rename (hope that's OK, if not, let me know, i'll drop it) and the other to show the ensemble name
- when hovering over the actual box of the box plots and
- in the legend

still need to test with several ensembles, ideally... i'll try to figure it (how to produce several ensembles) out myself in the meanwhile

valentin-krasontovitsch:
hmm managed to get another ensemble in there (running `ert`'s `ensemble_experiment` and `ensemble_smoother` with target-case `smoother` - is this the/a way?) and it seems like the way i got the ensemble in there now (basically hijacking / using the `name` property of each boxplot), the plots for same dates are no longer grouped together!

misfits-two-ensembles-ungrouped

I suppose that is not desired, eh?

valentin-krasontovitsch:
alrighty i got something which might work now

- box plots are overlayed again
- ensemble name is shown in box hover text
- legend displays ensemble

here is new stuff which might be meh:
- the legends are grouping ensembles together, based on (in my case) date
- the date is formatted on the x-axis to just a year
- the date is interpreted / recognized as a date, and thus the x-axis becomes linear time rather than categorical, as it perhaps was before

are any of these a problem? if yes, i can try to overcome.

misfits-two-ensembles-grouped

oysteoh:
> hmm managed to get another ensemble in there (running `ert`'s `ensemble_experiment` and `ensemble_smoother` with target-case `smoother` - is this the/a way?) and it seems like the way i got the ensemble in there now (basically hijacking / using the `name` property of each boxplot), the plots for same dates are no longer grouped together!
> 
> misfits-two-ensembles-ungrouped
> 
> I suppose that is not desired, eh?

The easiest thing is to use the already ran snake case which have a lot of data. 

`cd ert/test-data/local/snake_oil/snake_oil.ert`
`ert vis snake_oil.ert`

oysteoh:
Just havent had time today as i have been occupied with user support and failing komodo bleeding . Will have a look as soon as possible - maybe @DanSava or @frode-aarstad also could have a look ?

valentin-krasontovitsch:
sure thing, no problem, and thank you : )

valentin-krasontovitsch:
one more note i should add - the legend grouping implies that it is no longer possible to isolate one ensemble; one can only isolate a (in my case) date, and then see all ensembles for the chosen date. or isolate several, or instead of isolate, hide one or several dates.

oysteoh:
Think this looks ok now - could  you please rebase and squash to make it possible to test with an up-to-date PR ? :+1:

valentin-krasontovitsch:
great, thanks! I rebased and fixed up, and pushed - and both tests have run and succeeded.

valentin-krasontovitsch:
thank you : )

PR #308, opened 2022-07-06
test-concept: Redesign search fields with buttons, put buttons on top

valentin-krasontovitsch:
just a show case


valentin-krasontovitsch:
closing in favor of original request #260

PR #309, opened 2022-07-07
Add a beta label

### LINKED ISSUE #248: Labelling pages with "beta"
We should have the ability to mark pages released in our stable environment with `beta` or equivalent in order for us to communicate to our users this is a new feature which should not be expected to be completed ( and also maybe that we especially wants feedback on it )

frode-aarstad:
**Issue**
Resolves #248 

## Pre review checklist

### DIFF
--- a/webviz_ert/plugins/_observation_analyzer.py
+++ b/webviz_ert/plugins/_observation_analyzer.py
@@ -1,30 +1,41 @@
 import dash
+import webviz_ert.models
+import webviz_ert.controllers
 from dash.development.base_component import Component
 from typing import List, Dict
-from dash import html
-
-import webviz_ert.models
-
 from webviz_ert.views import ensemble_selector_list, response_obs_view
-
 from webviz_ert.plugins import WebvizErtPluginABC
-import webviz_ert.controllers
 
 
 class ObservationAnalyzer(WebvizErtPluginABC):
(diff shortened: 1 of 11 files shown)


webviz_ert/plugins/_observation_analyzer.py
```diff
@@ -12,14 +12,21 @@
...
+    def __init__(self, app: dash.Dash, project_identifier: str, beta: bool = False):
         super().__init__(app, project_identifier)
         self.set_callbacks(app)
+        self.beta = beta
 
     @property
     def layout(self) -> Component:
         return html.Div(
             [
+                html.Div(
```
oysteoh, line 24: I wonder if we should do something more to this text - feels a bit strange only heaving BETA as header... 🤔  Would it make sense having a descriptive text instead / in addition? 
```
* This page is considered a beta version and could be changed or removed. You are encouraged to use it and give feedback to us regarding functionality and / or bugs. 
```

oysteoh:
Looks good ! One minor comment regarding the text and how we are visualizing this is a beta to make it more clear for the users.

oysteoh (approved):
I think this looks good now ! Informative message to the user 👍 👍  I'll approve but we should get the jenkins test to also pass - that is currently not possible but will be as soon as bleeding is rebuilt.

frode-aarstad:
Jenkins test this please

PR #310, opened 2022-07-08
Set first obs as default value in correlation view

### LINKED ISSUE #301: Response correlation viewer - First observation should be the initial selection
feedback2

hnformentin:
**Issue**
Resolves #301

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -2,7 +2,7 @@ import pandas as pd
 import plotly.graph_objects as go
 import dash
 
-from typing import List, Optional, Dict, Tuple, Any
+from typing import List, Optional, Dict, Tuple, Any, Union
 from copy import deepcopy
 from dash import dcc, html
 from dash.development.base_component import Component
@@ -162,8 +162,8 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         if not (ensembles and responses and corr_param_resp["response"] in responses):
             raise PreventUpdate
         selected_response = corr_param_resp["response"]
-        _plots = []
-        _obs_plots: List[PlotModel] = []
+        plots = []
+        obs_plots: List[PlotModel] = []
 
         loaded_ensembles = [
             load_ensemble(parent, ensemble_id) for ensemble_id in ensembles
@@ -173,8 +173,10 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             response = ensemble.responses[selected_response]
 
             x_axis = response.axis
+
             if isinstance(x_axis, pd.Index) and x_axis.empty:
                 continue
+
             if x_axis is not None:
                 if str(x_axis[0]).isnumeric():
                     style = deepcopy(assets.ERTSTYLE["response-plot"]["response-index"])
@@ -185,7 +187,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             style.update({"marker": {"color": ensemble_color}})
             style.update({"line": {"color": ensemble_color}})
 
-            _plots += [
+            plots += [
                 PlotModel(
                     x_axis=x_axis,
                     y_axis=data_df[realization],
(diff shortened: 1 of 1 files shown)


hnformentin:
To do:
- check mypy errors
- add tests: should they be on the element of the interface (e.g. vertical line plotted) or in the callback function (`update_corr_param_resp` or `update_response_overview_plot`)
- verify way variables are being transformed

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -173,8 +173,15 @@ def update_response_overview_plot(
             response = ensemble.responses[selected_response]
 
             x_axis = response.axis
+
+            if x_axis is None:
```
hnformentin, line 177: This was needed because mypy complained that `None` could neither have `get_loc` nor `index` methods in line [213](https://github.com/equinor/webviz-ert/pull/310/files#diff-51ada3feb545708cc141a34f1482d91f92630bed64d5501567bf3eae09c43fddR213)
This will make the [None condition below](https://github.com/equinor/webviz-ert/pull/310/files#diff-51ada3feb545708cc141a34f1482d91f92630bed64d5501567bf3eae09c43fddL178) to be outdated. 
But as I am not sure about the if-statement above, I did not remove the below.

Which situation we can have `None` as data type to `x_axis`?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -232,6 +247,18 @@ def update_response_overview_plot(
...
+        :return: The first x value in the observation data, converted
+            to type suitable for lookup in the response vector.
+        """
+        if type(obs_data["x_axis"][0]) == str:
+            return int(obs_data["x_axis"][0])
+        elif type(obs_data["x_axis"][0]) == pd._libs.tslibs.timestamps.Timestamp:
+            return str(obs_data["x_axis"][0])
+        else:
+            return obs_data["x_axis"][0]
+
```
hnformentin, line 261: I am not comfortable with the need of this function. 
Can we try to get rid of the inconsistence between data type of the x axis from the observation and response?
Maybe it is something necessary or even a feature?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -374,7 +401,11 @@ def update_corr_index(
         triggered_id = ctx.triggered[0]["prop_id"].split(".")[0]
 
         if triggered_id == parent.uuid("parameter-selection-store-resp"):
-            return {response: corr_xindex.get(response, 0) for response in responses}
+            default_index = 0
```
hnformentin, line 404: I tried to let explicit that we have a default value because it took me more than a second to understand how it worked :D

hnformentin:
Related to tests: should they be on the element of the interface (e.g. vertical line plotted inspected on the webpage) or in the callback function (`update_corr_param_resp` or `update_response_overview_plot`?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -173,8 +173,15 @@ def update_response_overview_plot(
...
 
             x_axis = response.axis
+
+            if x_axis is None:
+                raise ValueError(
+                    f"No x_axis label in the response {selected_response}."
+                )
+
             if isinstance(x_axis, pd.Index) and x_axis.empty:
                 continue
```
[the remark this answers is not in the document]
oysteoh: If i remember correctly - we can have keys which doesnt have data for certain ensembles. I guess this relates to that situation. Instead of crashing we want to just display an empty chart.

hnformentin:
Test this please!

hnformentin:
Test this please!

webviz_ert/controllers/response_correlation_controller.py
hnformentin: I see, so it is not inconsistency if we need both. Eventually I will understand why :D

hnformentin:
It is working for snake_oil test, where the two first plots have observations and the last one does not:
Screenshot 2022-07-13 at 07 46 23
Screenshot 2022-07-13 at 07 47 01
Screenshot 2022-07-13 at 07 47 18

webviz_ert/controllers/response_correlation_controller.py
frode-aarstad: Perhaps prefix function with _get.. if its internal to this class and do a Union of the possible return types

frode-aarstad: That could be a good idea if its an error if the datatype is not str or timestamp.

hnformentin:
@frode-aarstad  I understood in our chat that I should remove `_` from variable names e.g. `_plots` and `_obs_plots` because this is outdated formatting. Is it correct?

frode-aarstad:
> @frode-aarstad I understood in our chat that I should remove `_` from variable names e.g. `_plots` and `_obs_plots` because this is outdated formatting. Is it correct?

Yes and perhaps also _layout

webviz_ert/controllers/response_correlation_controller.py
hnformentin: ok, a note is that `get_first_observation_x` is part of the function `response_correlation_controller` and not part of a class. But I guess it is still valid your suggestion.

hnformentin:
@frode-aarstad I will rebase the last commit into the previous one when approved and before merging :-)

frode-aarstad (approved):
Looks good 🚀

hnformentin:
Thanks !

PR #322, opened 2022-07-18
Response Correlation: Ordering of parameters, hover text visibility

### LINKED ISSUE #304: Response correlation viewer - Tornado plot ordering
Heatmap and tornado plot automatically order by weak -> strong correlation and makes it hard to keep track of what is going on when clicking around. Should be possible to toggle on/off this ordering. Pop-up box showing info when hover the heatmap gets in the way. ~~Add column names.~~
feedback6

valentin-krasontovitsch:
**Issue**
Closes #304 

**Approach**
See commit messages.

I've made several small (pairwise independent) commits with detailed messages - I'm not sure what is the best approach, gonna leave that up for discussion during review. I've implemented several of the ideas mentioned in the issue, instead of waiting for feedback.

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -39,6 +39,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             Input(parent.uuid("correlation-store-xindex"), "data"),
             Input(parent.uuid("correlation-store-selection"), "data"),
             Input(parent.uuid("correlation-metric"), "value"),
+            Input(parent.uuid("sort-parameters"), "on"),
         ],
         [
             State(parent.uuid("parameter-selection-store-param"), "data"),
@@ -50,6 +51,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         corr_xindex: Dict,
         corr_param_resp: Dict,
         correlation_metric: str,
+        sort_parameters: bool,
         parameters: List[str],
         responses: List[str],
         ensembles: List[str],
@@ -86,9 +88,8 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             corrdf = corrdf.drop(valid_responses, axis=0).fillna(0)
             if selected_response not in valid_responses:
                 continue
-            if df_index is None:
-                df_index = corrdf[selected_response].abs().sort_values().index.copy()
-            corrdf = corrdf.reindex(df_index)
+            if sort_parameters:
+                corrdf, df_index = sort_dataframe(corrdf, df_index, selected_response)
             # create heatmap
             data_heatmap = {
                 "type": "heatmap",
@@ -458,6 +459,17 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
 
         parent.save_state("active_correlation", corr_param_resp)
         return corr_param_resp
+
+
+def sort_dataframe(
+    dataframe: pd.core.frame.DataFrame,
+    index: Optional[pd.core.indexes.base.Index],
+    sort_by_column: str,
+) -> Tuple[pd.core.frame.DataFrame, pd.core.indexes.base.Index]:
+    if index is None:
+        index = dataframe[sort_by_column].abs().sort_values().index.copy()
+    dataframe = dataframe.reindex(index)
+    return dataframe, index
 
 
 def _define_style_ensemble(index: int, x_axis: pd.DataFrame) -> Dict:
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -40,6 +40,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             Input(parent.uuid("correlation-store-selection"), "data"),
             Input(parent.uuid("correlation-metric"), "value"),
             Input(parent.uuid("sort-parameters"), "on"),
+            Input(parent.uuid("hide-hover"), "on"),
         ],
         [
             State(parent.uuid("parameter-selection-store-param"), "data"),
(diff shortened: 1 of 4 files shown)


webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -87,7 +91,12 @@ def update_correlation_plot(
             if selected_response not in valid_responses:
                 continue
             if df_index is None:
-                df_index = corrdf[selected_response].abs().sort_values().index.copy()
+                if sort_parameters:
+                    df_index = (
+                        corrdf[selected_response].abs().sort_values().index.copy()
+                    )
+                else:
+                    df_index = corrdf[selected_response].abs().index.copy()
```
hnformentin, line 99: Would it be possible to isolate this functionality in a new function and test it?

hnformentin:
Quite an elegant solution 👏 also, nice commit messages.
Some print screens of the changes:

Before: 
Screenshot 2022-07-19 at 07 45 12

After:
Screenshot 2022-07-19 at 07 44 16
Screenshot 2022-07-19 at 07 44 36

valentin-krasontovitsch:
thank you ^^ and thanks for adding the screen shots! I should have done it myself, shouldn't be lazy...

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: you know, that's a very good question. `df_index` only gets used in line 100, to reindex the existing correlation data frame. and perhaps (I'm not sure) the reindexing existed in the code only because the data was getting sorted! so if we don't sort, we don't need to reindex, and we don't need `df_index` at all!
gonna try this out : )

not sure about the own function and testing it - i'm gonna think a bit about that one, see that I can pull something out and write a meaningful test :sweat_smile:

valentin-krasontovitsch: it seems like you were right - we only needed the index when we have to sort. i've adjusted the code accordingly.

about the test and refactoring - I don't know if this is what you meant, or if it adds much value, but the last commit extracts the part of the code responsible for sorting the dataframe, and i've added a test for testing said function. let me know what you think : )

tests/controllers/test_response_correlation.py
```diff
@@ -0,0 +1,18 @@
...
+def test_sort_dataframe():
+    one_key = "WOPR"
+    other_key = "BGMC"
+    data = {}
+    data[one_key] = [0,3,2]
+    data[other_key] = [8,21,2]
+    dataframe = pd.DataFrame(data=data)
+    index = None
+    sorted_dataframe, _ = sort_dataframe(dataframe, index, other_key)
+    print(sorted_dataframe)
```
hnformentin: Remove print...

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -23,6 +23,17 @@
...
 
+def sort_dataframe(
+    dataframe: pd.core.frame.DataFrame,
+    index: Optional[pd.core.indexes.base.Index],
+    sort_by_column: str,
+) -> Tuple[pd.core.frame.DataFrame, pd.core.indexes.base.Index]:
+    if index is None:
+        index = dataframe[sort_by_column].abs().sort_values().index.copy()
+    dataframe = dataframe.reindex(index)
+    return dataframe, index
```
hnformentin, line 34: Regarding vertical order, we expect the most important concepts to come first, so the natural order would be to put it after `response_correlation_controller`.

tests/controllers/test_response_correlation.py
```diff
@@ -0,0 +1,18 @@
...
+from webviz_ert.controllers.response_correlation_controller import (
+    sort_dataframe,
+)
+
+def test_sort_dataframe():
+    one_key = "WOPR"
+    other_key = "BGMC"
+    data = {}
+    data[one_key] = [0,3,2]
+    data[other_key] = [8,21,2]
```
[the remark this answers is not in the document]
valentin-krasontovitsch: oh yes you're definitely right, that makes a lot of sense : )
i'll change the values.

tests/controllers/test_response_correlation.py
valentin-krasontovitsch: oh thank you, will do :sweat_smile:

hnformentin:
Interesting that you created a new file for tests `tests>controllers>test_response_correlation_controller.py` while the other tests were in `tests>views>test_response_correlation.py`. reading what we have in the corresponding folders in `webviz_ert`, I think that it makes a lot of sense what you have done. But then, I think we should move all tests below the line (141)[https://github.com/valentin-krasontovitsch/webviz-ert/blob/resp-cor-view-tornado-ordering/tests/views/test_response_correlation.py#L141] to `tests>controllers>test_response_correlation_controller.py` . What do you think? If this is to be done, not to do in this issue.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: good to know, I'll move it down! is that a python style convention, scout convention, equinor convention, or just basic understanding? :innocent: 
my personal intuition is to put helper / utility functions like this one above the "main" or important functions... but that's just me ^^

valentin-krasontovitsch:
yeah in general the dash design in this project tries to separate concerns, by using views and controllers, but my feeling is that a clean separation cannot be achieved ("view" elements are determined in the controller, like what the plots look like) - this might be based on an unprecise definition (in my head) of what belongs to views, and what to controllers.

anyways, yes, sure, after your refactoring efforts, the tests below line 141 in `tests>views>test_response_correlation.py` are clearly tests of (utility functions of) the controller, and hence can be moved to the `controllers` folder : )

tests/controllers/test_response_correlation.py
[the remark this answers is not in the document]
valentin-krasontovitsch: updated to range -1 to 1!

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: moved it down!

tests/controllers/test_response_correlation.py
valentin-krasontovitsch: print is gone : )

hnformentin (approved):
LGTM!

PR #325, opened 2022-07-19
Fix bug of moving selection of vertical line

### LINKED ISSUE #324: Fix vertical line of response in correlation view
**Describe the bug**
The vertical line is not moving when another point in the response graph is clicked. 

**To Reproduce**
1. Click on correlation view
2. Select an ensemble, a response and a parameter
3. Click in a different point in the response graphic
4. Verify that the vertical line does not move

**Expected behavior**
The vertical line should move as selected by clicking.

hnformentin:
**Issue**
Resolves #324

**Approach**
The problem was only observed for responses with observation.
The solution was to add an if statement before updating where the vertical line should be displayed.
The problem of this solution is that it wont be possible to select the index 0.
I don't see this as a problem, as the request of the user indicated that the index 0 does aggregate value to the analysis.

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -213,17 +213,18 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
                 for obs in response.observations:
                     obs_plots.append(_get_observation_plots(obs.data_df()))
 
-                x_axis_default_observation = _get_first_observation_x(
-                    response.observations[0].data_df()
-                )
-                if isinstance(x_axis, pd.Index):
-                    corr_xindex[selected_response] = x_axis.get_loc(
-                        x_axis_default_observation
+                if corr_xindex[selected_response] == 0:
+                    x_axis_default_observation = _get_first_observation_x(
+                        response.observations[0].data_df()
                     )
-                elif isinstance(x_axis, list):
-                    corr_xindex[selected_response] = x_axis.index(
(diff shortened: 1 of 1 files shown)


hnformentin:
Now the line is displayed upon selection.
Screenshot 2022-07-19 at 11 25 05
Screenshot 2022-07-19 at 11 25 16

hnformentin:
Test this please!

DanSava (approved):

valentin-krasontovitsch:
just a quick check - maybe i'm doing something wrong. could you try to, with either `main` or the head of this pr,

- open webviz-ert response correlation viewer
- click one ensemble, one response, one paramter (the parameter is optional)
- reload the browser tab with F5 / ctrl-r

i am observing that the upper left plot is drawn before reload, but remains blank after reload O_o

also in other news this is really unrelated to this PR... sorry about hijacking it O:)

valentin-krasontovitsch (approved):
yeah I'm looking at the code, and I don't see a better way of checking for first time draw, either. I think there might be a way, but right now the state and how everything is connected is just too complicated for me to wrap my head around quickly. i think we can leave it for later, when things are a bit more refactored ; )

hnformentin:
> yeah I'm looking at the code, and I don't see a better way of checking for first time draw, either. I think there might be a way, but right now the state and how everything is connected is just too complicated for me to wrap my head around quickly. i think we can leave it for later, when things are a bit more refactored ; )

Same behaviour here. Created an issue to deal with it.

PR #331, opened 2022-07-20
Refactor corr index

### LINKED ISSUE #323: Refactor update_response_overview_plot
Continue refactoring update_response_overview_plot

hnformentin:
- Change order tests coherent with functions/folders
- Extract _update_corr_index_dict and add test

**Issue**
Resolves #323

### DIFF
--- a/tests/controllers/test_response_correlation_controller.py
+++ b/tests/controllers/test_response_correlation_controller.py
@@ -1,4 +1,10 @@
 import pandas as pd
+import pytest
+from webviz_ert.controllers.response_correlation_controller import (
+    _get_first_observation_x,
+    _define_style_ensemble,
+    _update_corr_index_dict,
+)
 
 from webviz_ert.controllers.response_correlation_controller import (
     sort_dataframe,
@@ -16,3 +22,78 @@ def test_sort_dataframe():
     sorted_dataframe, _ = sort_dataframe(dataframe, index, other_key)
     assert list(sorted_dataframe[other_key]) == sorted(data[other_key])
     assert list(sorted_dataframe[one_key]) == [0.2, -0.4, 0.6]
+
+
+@pytest.mark.parametrize(
+    "index,expected_color", [(0, "rgba(56,108,176,0.8)"), (1, "rgba(127,201,127,0.8)")]
+)
+def test_define_style_ensemble_color(index, expected_color):
+    x_axis = pd.DataFrame([1])
+    style = _define_style_ensemble(index, x_axis)
+    assert style["line"]["color"] == expected_color
+    assert style["marker"]["color"] == expected_color
+
+
+@pytest.mark.parametrize(
+    "x_axis_list,expected_mode",
+    [
+        (
+            [pd._libs.tslibs.timestamps.Timestamp("01-01-2020")],
+            "markers+lines",
+        ),
+        ([str(1)], "markers"),
+    ],
+)
+def test_define_style_ensemble_mode(x_axis_list, expected_mode):
+    x_axis = pd.Index(x_axis_list)
+    style = _define_style_ensemble(0, x_axis)
+    assert style["mode"] == expected_mode
+
+
+@pytest.mark.parametrize(
+    "response_x_axis,observation_x_axis_list,expected_index",
+    [
+        (
+            [
+                "2020-01-01 00:00:00",
+                "2020-01-07 00:00:00",
+                "2020-01-14 00:00:00",
+                "2020-01-21 00:00:00",
+            ],
+            [
+                pd._libs.tslibs.timestamps.Timestamp("14-01-2020"),
+                pd._libs.tslibs.timestamps.Timestamp("01-02-2020"),
+            ],
+            2,
+        ),
+        (pd.Index([0, 1]), [str(1)], 1),
+    ],
+)
+def test_update_corr_index_dict(
+    response_x_axis, observation_x_axis_list, expected_index
+):
+    observation_x_axis = pd.DataFrame(observation_x_axis_list, columns=["x_axis"])
+    updated_index = _update_corr_index_dict(response_x_axis, observation_x_axis)
+    assert expected_index == updated_index
+
+
+@pytest.mark.parametrize(
+    "observation,expected",
+    [
+        ([str(1)], int(1)),
+        (
+            [pd._libs.tslibs.timestamps.Timestamp("01-01-2020")],
+            str("2020-01-01 00:00:00"),
+        ),
+    ],
+)
+def test_get_first_observation_x_valid(observation, expected):
+    df_observation = pd.DataFrame(observation, columns=["x_axis"])
+    result = _get_first_observation_x(df_observation)
+    assert result == expected
+
+
+def test_get_first_observation_x_invalid():
+    df_observation = pd.DataFrame([int(1)], columns=["x_axis"])
+    with pytest.raises(ValueError, match="invalid obs_data type"):
+        _get_first_observation_x(df_observation)
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -209,7 +209,8 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             if isinstance(x_axis, pd.Index) and x_axis.empty:
                 continue
 
-            style = _define_style_ensemble(index, x_axis)
+            if x_axis is not None:
+                style = _define_style_ensemble(index, x_axis)
 
             data_df = response.data_df().copy()
             plots += [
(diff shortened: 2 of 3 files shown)


tests/controllers/test_response_correlation_controller.py
```diff
@@ -16,3 +22,75 @@ def test_sort_dataframe():
...
+    [
+        ([pd._libs.tslibs.timestamps.Timestamp("01-01-2020")], "markers+lines"),
+        ([str(1)], "markers"),
+    ],
+)
+def test_define_style_ensemble_mode(x_axis_element, expected_mode):
+    x_axis = pd.Index(x_axis_element)
+    style = _define_style_ensemble(0, x_axis)
+    assert style["mode"] == expected_mode
+
```
sregales-TNO, line 48: It would be nice to have a test for x_axis = None.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -498,6 +488,24 @@ def _define_style_ensemble(index: int, x_axis: pd.DataFrame) -> Dict:
     return style
```
sregales-TNO, line 488: I know this is not part of the PR, but it is tested in this PR.

I see here that `x_axis` is not set as `Optional`, yet there is a conditional for to ensure it is not `None` that seems redundant.

Furthermore, `style` variable is only assigned if `x_axis` is not `None`. So what is the behaviour of `style.update` update if `x_axis` is `None`? is there a global `style` that I'm unaware of?

suggestion:
```python
style = deepcopy(assets.ERTSTYLE["response-plot"][f"response{'-index' if str(x_axis[0]).isnumeric() else ''}"])
# --------or for readability---------
label = '-index' if str(x_axis[0]).isnumeric() else ''
style = deepcopy(assets.ERTSTYLE["response-plot"][f"response{label}"])
# --------rest of the code---------
ensemble_color = assets.get_color(index=index)
.
.
.
```

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -498,6 +488,24 @@ def _define_style_ensemble(index: int, x_axis: pd.DataFrame) -> Dict:
...
 
 
+def _update_corr_index_dict(
+    response_x_axis: Union[pd.Index, list], observation_x_axis: pd.Index
```
sregales-TNO, line 492: We say that we expect `response_x_axis` to be either a `pd.Index` or `list`, yet later we check if parameter is one of the two. What would happen if user/developer calles function using neither options? `updated_index` is not even assigned in that case. This is me, but I tend to use `if` conditions for expects that are not part of the core functionality of the function. This will reduce (nested) `if` conditions, reduce the complexity, reduce chances of unassignment and as a side effect make the code more readable.
example:
```python
def _update_corr_index_dict(
    response_x_axis: Union[pd.Index, list], observation_x_axis: pd.Index
) -> int:
    if type(response_x_axis) not in (pd.Index, list):
        raise ValueError
    return {pd.Index: responds_x_axis.get_loc, list: response_x_axis.index}.get(response_x_axis)(_get_first_observation_x(observation_x_axis))
#--------readability----------
func_selector = {pd.Index: responds_x_axis.get_loc, list: response_x_axis.index}
return func_selector.get(response_x_axis)(x_axis_default_observation)
```

tests/controllers/test_response_correlation_controller.py
```diff
@@ -16,3 +22,75 @@ def test_sort_dataframe():
...
+            [
+                pd._libs.tslibs.timestamps.Timestamp("14-01-2020"),
+                pd._libs.tslibs.timestamps.Timestamp("01-02-2020"),
+            ],
+            2,
+        ),
+        (pd.Index([0, 1]), [str(1)], 1),
+    ],
+)
+def test_update_corr_index_dict(
```
sregales-TNO, line 69: Test for invalid parameters or use hypotheses

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -498,6 +488,24 @@ def _define_style_ensemble(index: int, x_axis: pd.DataFrame) -> Dict:
...
+    return updated_index
+
+
+def _layout_figure() -> dict:
+    layout = assets.ERTSTYLE["figure"]["layout"].copy()
+    layout.update(dict(showlegend=False))
+    return layout
+
+
 def _get_first_observation_x(obs_data: pd.DataFrame) -> Union[int, str]:
```
sregales-TNO, line 509: again not part of the PR, but part of the test sweep and can be updated. It be funny if all this is tests created prior to start refactoring.

deal with exeption to the core function first. then you can emplement the the function afterwards with ease. 
And this function can also use dict (my prefrence over multiple `if-elif-else` statments) to choose what to return.
no need for `if-elif-else` if we are returning after each condition.
suggestion:
```python
data_point = obs_data["x_axis"][0]
caster = {str: int, pd._libs.tslibs.timestamps.Timestamp: str}
if type(data_point) not in caster.keys():
    raise ValueError
return caster.get(type(data_point))(data_point)
```

sregales-TNO (commented):
Oh my, I went a bit overboard with the review. I did the review while in a meeting and now that I read the description I see that it is only moving the tests, sorry

webviz_ert/controllers/response_correlation_controller.py
DanSava: Cool suggestion @sregales-TNO but I am not sure I agree with everything suggested. 

I agree that adding a check on the expected types for the response_x_axis is a good idea with raising an error or implementing specific handling when we get unexpected types for `response_x_axis`. 

But I don't encourage using a map indexed on data types, looks a bit too funcky for my taist. 

also the suggested solution does not seem to work as it is, it requires replacing `response_x_axis` with `type(response_x_axis)`:

``` python
x_axis_default_observation = _get_first_observation_x(observation_x_axis) func_selector = {pd.Index: responds_x_axis.get_loc, list: response_x_axis.index} return func_selector.get(type(response_x_axis))(x_axis_default_observation)
```
and in my opinion, this is no more readable than just than what it was there before :shrug:

sregales-TNO: thanks for the correction. yeah, I guess it's a matter of preference, I normally swap to mappings when it's multiple consecutive conditional statements on the same element/object with slightly different behavior. I guess for two conditionals it seems overkill. A better use would be 3+ consecutive ones

tests/controllers/test_response_correlation_controller.py
hnformentin: Good point! While thinking to test the `None` condition I got the insight that this condition should not be encapsulated inside the `define_style_ensemble` function for now. There are multiple conditional tests in the main function that we don't fully understand (e.g. lines 209 and 250), and it seems better to let them explicit in the main function until we find out its final need. Also, I don't have any test case to understand what we expect when `x_axis == None`. 

Are the changes proposed ok?

webviz_ert/controllers/response_correlation_controller.py
hnformentin: Good points! Thanks. I think I addressed in the previous comment. If you agree, I will treat this checking conditions when refactoring the main function.

tests/controllers/test_response_correlation_controller.py
hnformentin: It is addressed in the comments above, right?

webviz_ert/controllers/response_correlation_controller.py
hnformentin: Nice suggestion! I copied pasted shamelessly, only with a minor change to make mypy happy :-) Thanks!

sregales-TNO (approved):
LGTM

hnformentin:
All tests pass locally, not on Jenkins. For recording, this is the log:

> =================================== FAILURES ===================================
> ___________________________ test_show_respo_with_obs ___________________________
> 
> mock_data = None
> dash_duo = <dash.testing.composite.DashComposite object at 0x7f0039afcca0>
> 
>     def test_show_respo_with_obs(mock_data, dash_duo):
>         """Test response observation filter works as expected"""
>         plugin = setup_plugin(dash_duo, __name__, ResponseCorrelation)
>     
>         select_ensemble(dash_duo, plugin, "default3")
>     
>         resp_selector_id = plugin.uuid("parameter-selector-multi-resp")
>         default_responses = get_options(dash_duo, "#" + resp_selector_id)
>     
>         expected_def_responses = ["FGPT", "FOPR", "SNAKE_OIL_GPR_DIFF", "WOPR:OP1"]
>         assert default_responses == expected_def_responses
>     
>         expected_obs_responses = ["FOPR", "WOPR:OP1"]
>         obs_radio_btn = dash_duo.find_element(
>             "#" + plugin.uuid("response-observations-check")
>         )
>         obs_radio_btn.click()
>         obs_responses = get_options(dash_duo, "#" + resp_selector_id)
> >       assert obs_responses == expected_obs_responses
> E       AssertionError: assert ['FGPT', 'FOP...', 'WOPR:OP1'] == ['FOPR', 'WOPR:OP1']
> E         At index 0 diff: 'FGPT' != 'FOPR'
> E         Left contains 2 more items, first extra item: 'SNAKE_OIL_GPR_DIFF'
> E         Full diff:
> E         - ['FOPR', 'WOPR:OP1']
> E         + ['FGPT', 'FOPR', 'SNAKE_OIL_GPR_DIFF', 'WOPR:OP1']
> 
> tests/views/test_response_correlation.py:133: AssertionError

hnformentin:
Test this please!

PR #332, opened 2022-07-22
New obs selector controller

### LINKED ISSUE #300: Response correlation viewer - New selector
We decided to make a new selector which visualizes all observations in the selected responses - something like this. The selected responses should be assigned a color when selected and the observations with the same color plotted along one ( or two depending on unit ) x-axis together with the rest of the selected responses. It should be possible to select multiple observations across the different  responses - but we should restrict to some limit which makes sense to avoid extensive load times.

Image from iOS

### LINKED ISSUE #302: Response correlation viewer - All observations in the selected responses should be available in the heatmap
feedback3

DanSava:
**Issue**
Resolves #300 

After a bit of thought decided to resolve #302 together with this one

## Pre-review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -1,6 +1,9 @@
+import numpy
 import pandas as pd
+from pandas._libs.tslibs.timestamps import Timestamp
 import plotly.graph_objects as go
 import dash
+import dateutil.parser
 
 from typing import List, Optional, Dict, Tuple, Any, Union
 from copy import deepcopy
@@ -9,18 +12,41 @@ from dash.development.base_component import Component
 from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output, State
 from plotly.subplots import make_subplots
+from plotly.graph_objs.layout import XAxis
 
 from webviz_ert.plugins import WebvizErtPluginABC
-from webviz_ert.models import (
-    load_ensemble,
-    BarChartPlotModel,
-    PlotModel,
-)
+from webviz_ert.models import load_ensemble, BarChartPlotModel, PlotModel, AxisType
 from webviz_ert.controllers.multi_response_controller import (
     _get_observation_plots,
     axis_label_for_ensemble_response,
 )
 from webviz_ert import assets
+
+
+def _get_selected_indexes(
+    plots: List[PlotModel], selected_data: Optional[Dict]
+) -> Dict:
+    selected_indexes: Dict = {}
+    if not selected_data:
+        return selected_indexes
+    for plot in plots:
+        selected_indexes[plot.name] = []
+        x_data_range = selected_data["range"].get("x")
+        x2_data_range = selected_data["range"].get("x2")
+        if plot.axis_type == AxisType.INDEX:
+            if not x2_data_range:
+                continue
+            x_start, x_end = x2_data_range
+        elif not x_data_range:
+            continue
+        else:
+            x_start = dateutil.parser.isoparse(x_data_range[0])
+            x_end = dateutil.parser.isoparse(x_data_range[1])
+
+        for val in plot._x_axis:
+            if x_start <= val <= x_end:
+                selected_indexes[plot.name].append(val)
+    return selected_indexes
 
 
 def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) -> None:
@@ -36,7 +62,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             ),
         ],
         [
-            Input(parent.uuid("correlation-store-xindex"), "data"),
+            Input(parent.uuid("correlation-store-selected-obs"), "data"),
             Input(parent.uuid("correlation-store-selection"), "data"),
             Input(parent.uuid("correlation-metric"), "value"),
             Input(parent.uuid("sort-parameters"), "on"),
@@ -49,7 +75,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         ],
     )
     def update_correlation_plot(
-        corr_xindex: Dict,
+        selected_obs: Dict,
         corr_param_resp: Dict,
         correlation_metric: str,
         sort_parameters: bool,
@@ -65,7 +91,6 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             and corr_param_resp["response"] in responses
         ):
             raise PreventUpdate
-
         selected_response = corr_param_resp["response"]
         data = {}
         colors = {}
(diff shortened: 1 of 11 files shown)


hnformentin:
Please test this!

hnformentin:
Jenkins is flaky again, the log is:

`=================================== FAILURES ===================================
_______________________ test_webviz_response_correlation _______________________

dash_duo = <dash.testing.composite.DashComposite object at 0x7f01456bf8b0>

    @pytest.mark.spe1
    def test_webviz_response_correlation(dash_duo):
        plugin = setup_plugin(dash_duo, __name__, ResponseCorrelation)
    
        # Wait for the ensemble selector to be initialized
        dash_duo.wait_for_contains_text(
            "#" + plugin.uuid("ensemble-multi-selector"),
            "default",
            timeout=4,
        )
    
        ensemble_name = select_first(dash_duo, "#" + plugin.uuid("ensemble-multi-selector"))
        _verify_keys_in_menu(
            dash_duo, plugin, parameter_keys, "parameter-selector-multi-param"
        )
    
        _verify_keys_in_menu(
            dash_duo, plugin, response_keys, "parameter-selector-multi-resp"
        )
        response_name = select_first(
            dash_duo, "#" + plugin.uuid("parameter-selector-multi-resp")
        )
        param_name = select_first(
            dash_duo, "#" + plugin.uuid("parameter-selector-multi-param")
        )
>       dash_duo.wait_for_text_to_equal(
            "#" + plugin.uuid("response-info-text"),
            f"{response_name} @ 2015-02-01T00:00:00,\nparameter: {param_name}",
            30,
        )

tests/test_webviz_ert.py:69: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/prog/res/komodo/bleeding-py38-rhel7/root/lib/python3.8/site-packages/dash/testing/browser.py:322: in wait_for_text_to_equal
    return self._wait_for(
/prog/res/komodo/bleeding-py38-rhel7/root/lib/python3.8/site-packages/dash/testing/browser.py:262: in _wait_for
    return _wait.until(method(*args), msg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <selenium.webdriver.support.wait.WebDriverWait (session="4344e81d63ee26927f4c1ef42bfcbbab")>
method = <dash.testing.wait.text_to_equal object at 0x7f0145596c70>
message = 'text -> WGPT:PROD @ 2015-02-01T00:00:00,\nparameter: FIELD_PROPERTIES:POROSITY::0 not found within 30s'

    def until(self, method, message: str = ""):
        """Calls the method provided with the driver as an argument until the \
        return value does not evaluate to ``False``.
    
        :param method: callable(WebDriver)
        :param message: optional message for :exc:`TimeoutException`
        :returns: the result of the last call to `method`
        :raises: :exc:`selenium.common.exceptions.TimeoutException` if timeout occurs
        """
        screen = None
        stacktrace = None
    
        end_time = time.monotonic() + self._timeout
        while True:
            try:
                value = method(self._driver)
                if value:
                    return value
            except self._ignored_exceptions as exc:
                screen = getattr(exc, 'screen', None)
                stacktrace = getattr(exc, 'stacktrace', None)
            time.sleep(self._poll)
            if time.monotonic() > end_time:
                break
>       raise TimeoutException(message, screen, stacktrace)
E       selenium.common.exceptions.TimeoutException: Message: text -> WGPT:PROD @ 2015-02-01T00:00:00,
E       parameter: FIELD_PROPERTIES:POROSITY::0 not found within 30s

../../../../test-kenv/root/lib64/python3.8/site-packages/selenium/webdriver/support/wait.py:90: TimeoutException`

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -79,15 +106,28 @@ def update_correlation_plot(
                 continue
             valid_responses = []
             for response in responses:
-                response_df = ensemble.responses[response].data_df()
-                if response_df.empty:
+                response_el = ensemble.responses[response]
+                if response_el.empty:
```
hnformentin, line 110: What does `el` refer to in `response_el`?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -46,17 +72,19 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash)
             State(parent.uuid("parameter-selection-store-param"), "data"),
             State(parent.uuid("parameter-selection-store-resp"), "data"),
             State(parent.uuid("selected-ensemble-dropdown"), "value"),
+            State(parent.uuid("correlation-store-selected-obs"), "data"),
```
hnformentin, line 75: Why is the component property of `correlation-store-selected-obs` different in Input (as modified_timestamp) and State (as data)?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -79,15 +106,28 @@ def update_correlation_plot(
...
+                            ] = response_el.data.loc[resp_index]
+                            valid_responses.append(f"{response_el.name}::{idx}")
+                else:
+                    # TODO: consider removing this else clause not really
+                    #  any point is showing responses without observations but here for now
+                    parameter_df[response] = response_el.data.iloc[0]
+                    valid_responses.append(response)
             corrdf = parameter_df.corr(method=correlation_metric)
             corrdf = corrdf.drop(valid_responses, axis=0).fillna(0)
+            selected_response = f"{selected_response}::0"
```
hnformentin, line 130: I don't understand this. Can you maybe comment why we need this "transformation"?

hnformentin:
We discussed that it is a nice feature to have merged and test with the user. 

At the same time, we want to have cleaner code base. This will come later with additional PR's that @DanSava proposed to create. I suggest to have a PR related to refactoring and adding unit tests to the functions, particularly the callbacks `update_response_parameter_scatterplot` and `update_obs_selector_plot`. 

Also, I think it would not be a bad idea to let some comments on the controls that were added, so we could later have better view why we added `if` here and there. 

Otherwise I am testing locally, and we could have some screenshots displaying the feature implemented and the desired behaviour.

hnformentin (approved):

webviz_ert/controllers/response_correlation_controller.py
DanSava: `el` stands for element but I will refactor to replace response_el with response and response from ln 109 with response_name to be more specific

webviz_ert/controllers/response_correlation_controller.py
DanSava: This is a hack that will need to be fixed. Currently, it is needed because now that `valid_responses` are of the form `f"{response.name}::{idx}"` so it is impossible to check if the selected_response is part of the valid responses without this transformation.

I am sure this part of the code could be improved.

webviz_ert/controllers/response_correlation_controller.py
DanSava: This is a good question, there is  another places in the code where this pattern is used:

here:
```
    @app.callback(
        Output(
            parent.uuid("response-overview"),
            "figure",
        ),
        [
            Input(parent.uuid("correlation-store-xindex"), "modified_timestamp"),
            Input(parent.uuid("parameter-selection-store-resp"), "modified_timestamp"),
            Input(parent.uuid("ensemble-selection-store"), "modified_timestamp"),
            Input(parent.uuid("correlation-store-selection"), "modified_timestamp"),
            Input(parent.uuid("correlation-store-selected-obs"), "modified_timestamp"),
        ],
        [
            State(parent.uuid("parameter-selection-store-resp"), "data"),
            State(parent.uuid("ensemble-selection-store"), "data"),
            State(parent.uuid("correlation-store-xindex"), "data"),
            State(parent.uuid("correlation-store-selection"), "data"),
            State(parent.uuid("correlation-store-selected-obs"), "data"),
        ],
    )
    def update_response_overview_plot(
```

And I kind of assumed it was required because there is an issue with triggering the callback if we only had for example `Input(parent.uuid("correlation-store-selected-obs"), "data")` and we needed  `Input(parent.uuid("correlation-store-selected-obs"), "modified_timestamp")` to trigger the callback for every change in the state of parent.uuid("correlation-store-selected-obs"), but I never checked if that is true or not. 

I give it a test and see if I can eliminate the state and only use the input.

DanSava: gave it a quick test and it seems there is no obvious advantage to using the pattern:
```python
    Input(parent.uuid("correlation-store-selected-obs"), "modified_timestamp"),
    State(parent.uuid("parameter-selection-store-param"), "data"),
 ```
I just replaced it with 
```python
    Input(parent.uuid("correlation-store-selected-obs"), "data"),
```

Seems to work the same

DanSava:
jenkins test this

PR #334, opened 2022-07-26
Refactor layout and add test

### LINKED ISSUE #280: Refactor parameter_selector_controller
Eliminate the need for 

`parameter_type_store_id = parent.uuid(f"parameter-type-store-{suffix}")`

and 

```python 
dcc.Store(
    id=parent.uuid(f"parameter-type-store-{suffix}"),
    storage_type="session",
    data=data_type,
)
```

hnformentin:
**Issue**
Resolves #280

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -239,13 +239,11 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         for plot in plots:
             fig.add_trace(plot.repr)
 
-        fig.update_layout(_layout_figure())
-
         x_axis_label = axis_label_for_ensemble_response(
             loaded_ensembles[0], selected_response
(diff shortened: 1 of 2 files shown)


valentin-krasontovitsch (approved):
very nice! I see that most of the refactoring was due to me just adding on code in the main function instead of using the helper function. Sorry about that :sweat_smile:
Thanks for the continuing refactoring efforts ^^

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -502,9 +499,13 @@ def _update_corr_index_dict(
     return updated_index
 
 
-def _layout_figure() -> dict:
```
eivindjahren, line 505: It should be possible to define this closer to the use site by having it be inside `response_correlation_controller`. You  can still access it inside the test by calling
`response_correlation_controller._layout_figure("label")`.

PR #335, opened 2022-07-26
Rename variables in update_response_overview_plot for readability

### LINKED ISSUE #280: Refactor parameter_selector_controller

hnformentin:
**Issue**
Resolves #280

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -194,34 +194,34 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         if not (ensembles and responses and corr_param_resp["response"] in responses):
             raise PreventUpdate
         selected_response = corr_param_resp["response"]
-        plots = []
-        obs_plots: List[PlotModel] = []
-
         loaded_ensembles = [
             load_ensemble(parent, ensemble_id) for ensemble_id in ensembles
         ]
 
+        response_plots = []
+        obs_plots: List[PlotModel] = []
+
         for index, ensemble in enumerate(loaded_ensembles):
             response = ensemble.responses[selected_response]
 
-            x_axis = response.axis
+            response_x_axis = response.axis
 
-            if isinstance(x_axis, pd.Index) and x_axis.empty:
+            if isinstance(response_x_axis, pd.Index) and response_x_axis.empty:
                 continue
 
-            if x_axis is not None:
-                style = _define_style_ensemble(index, x_axis)
+            if response_x_axis is not None:
+                style = _define_style_ensemble(index, response_x_axis)
 
-            data_df = response.data_df().copy()
-            plots += [
+            response_df = response.data_df().copy()
+            response_plots += [
                 PlotModel(
-                    x_axis=x_axis,
-                    y_axis=data_df[realization],
+                    x_axis=response_x_axis,
+                    y_axis=response_df[realization],
                     text=realization,
                     name=realization,
                     **style,
                 )
-                for realization in data_df
+                for realization in response_df
             ]
 
             if response.observations:
@@ -229,14 +229,14 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
                     obs_plots.append(_get_observation_plots(obs.data_df()))
 
                 if corr_xindex[selected_response] == 0 and isinstance(
-                    x_axis, (pd.Index, list)
+                    response_x_axis, (pd.Index, list)
                 ):
                     corr_xindex[selected_response] = _update_corr_index_dict(
-                        x_axis, response.observations[0].data_df()
+                        response_x_axis, response.observations[0].data_df()
                     )
 
         fig = go.Figure()
-        for plot in plots:
+        for plot in response_plots:
             fig.add_trace(plot.repr)
 
         x_axis_label = axis_label_for_ensemble_response(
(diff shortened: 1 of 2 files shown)


webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -194,49 +194,49 @@ def update_response_overview_plot(
...
+
         for index, ensemble in enumerate(loaded_ensembles):
             response = ensemble.responses[selected_response]
 
-            x_axis = response.axis
+            response_x_axis = response.axis
 
-            if isinstance(x_axis, pd.Index) and x_axis.empty:
+            if isinstance(response_x_axis, pd.Index) and response_x_axis.empty:
                 continue
```
hnformentin, line 210: @DanSava Do you know why we need this `if` above, e.g. when this is the case and what would be the `else` statement?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -194,49 +194,49 @@ def update_response_overview_plot(
...
 
-            if x_axis is not None:
-                style = _define_style_ensemble(index, x_axis)
```
hnformentin, line 213: @DanSava Do you know in which context `x_axis´ would be `None`?

webviz_ert/controllers/response_correlation_controller.py
DanSava: It is needed to account for the cases where a response is empty (an empty dataframe or the index of the dataframe is not set correctly) this happens when data is corrupted in storage or there is no data output for that response.

>  what would be the else statement?

Not sure what you mean by this, but not all if statements require an `else` clause. In this case on the else branch, you would execute all the code below that is why we don't have any `else` here.

webviz_ert/controllers/response_correlation_controller.py
DanSava: That is a good question not really sure, looking at the logs it seems you added the check here, and I guess it is not really needed and can be removed given the check above line 209

DanSava: also just as a side thing I notice while looking at the code:

```python
def _define_style_ensemble(index: int, x_axis: pd.DataFrame) -> Dict:
    if str(x_axis[0]).isnumeric():
        style = deepcopy(assets.ERTSTYLE["response-plot"]["response-index"])
    else:
        style = deepcopy(assets.ERTSTYLE["response-plot"]["response"])
    ensemble_color = assets.get_color(index=index)
    style.update({"marker": {"color": ensemble_color}})
    style.update({"line": {"color": ensemble_color}})

    return style
```

`_define_style_ensemble` expects  `x_axis: pd.DataFrame` and style = _define_style_ensemble(index, x_axis) passes and x_axis of type `List[Union[int, str, datetime]]` so there is a type mismatch 


The same thing is true for `_get_first_observation_x`:
it expects a dataframe `def _get_first_observation_x(obs_data: pd.DataFrame) -> Union[int, str]:`
but when it is used it gets a `observation_x_axis` of type pd.Index 

So there is a bit of a mismatch there as well, not really sure how this escaped type-checking :shrug:

hnformentin: > That is a good question not really sure, looking at the logs it seems you added the check here, and I guess it is not really needed and can be removed given the check above line 209

Actually it was added before, and I inherited because did not know why it was needed: https://github.com/hnformentin/webviz-ert/blob/16df3f0e793a77fdd60ef6d6bd7d562f86a94ecb/webviz_ert/controllers/response_correlation_controller.py#L178

You could be right in no damage in removing it, I will check with some tests.

DanSava: > Actually it was added before, and I inherited because did not know why it was needed

You are right, sorry about that, I got confused because you removed it one PR and then you added back in another.

>I will check with some tests. 

:+1:

hnformentin: > also just as a side thing I notice while looking at the code:
> 
> ```python
> def _define_style_ensemble(index: int, x_axis: pd.DataFrame) -> Dict:
>     if str(x_axis[0]).isnumeric():
>         style = deepcopy(assets.ERTSTYLE["response-plot"]["response-index"])
>     else:
>         style = deepcopy(assets.ERTSTYLE["response-plot"]["response"])
>     ensemble_color = assets.get_color(index=index)
>     style.update({"marker": {"color": ensemble_color}})
>     style.update({"line": {"color": ensemble_color}})
> 
>     return style
> ```
> 
> `_define_style_ensemble` expects `x_axis: pd.DataFrame` and style = _define_style_ensemble(index, x_axis) passes and x_axis of type `List[Union[int, str, datetime]]` so there is a type mismatch
> 
> The same thing is true for `_get_first_observation_x`: it expects a dataframe `def _get_first_observation_x(obs_data: pd.DataFrame) -> Union[int, str]:` but when it is used it gets a `observation_x_axis` of type pd.Index x_axis_default_observation = _get_first_observation_x(observation_x_axis)
> 
> So there is a bit of a mismatch there as well, not really sure how this escaped type-checking 🤷

Good points!
For `_define_style_ensemble` the correct data_type seems to be `pd.Index`. I have the output `<class 'pandas.core.indexes.numeric.Int64Index'>` for `print(type(response_x_axis))` before calling `_define_style_ensemble` in line 215.

For `_get_first_observation_x`, the type should be pd.DataFrame, so I correct to have consistency as you pointed out.

DanSava (approved):
:+1:

PR #338, opened 2022-07-26
Fix ensemble handling in response correlation viewer

### LINKED ISSUE #326: Response Correlation Viewer: ensemble handling
**Describe the bug**
Not all plots get updated when one of several ensembles gets removed from selection.

**To Reproduce**
3. Choose at least two ensembles, some responses, and some parameters
4. Deselect one of the ensembles
5. Only the upper left plot gets updated, and data for the deselected ensemble is removed

**Expected behavior**
All plots get updateed, and parts corresponding to the deselected ensemble are removed

**Screenshots**
Two ensembles and several responses and parameters selected:

ensembl-issu1-before-deselect

Same situation, as above, but one ensemble deselected:

ensembl-issu1-after-deselect png

 - OS: Ubuntu 22.04 "jammy"
 - Browser: Google Chrome 102.0.5005.115

valentin-krasontovitsch:
**Issue**
Resolves #326

**Approach**
We
- use the store instead of html elements to fetch info about which ensembles are selected and
- return blank (graph) objects instead of raising a don't-update-exception

**Testing**
We thought about adding regression tests, but it turned out to be at least difficult - I could not see any immediate way of querying the DOM using the dash duo testing api to figure out whether a graph was drawn, or not.

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -90,7 +90,8 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             and responses
             and corr_param_resp["response"] in responses
         ):
-            raise PreventUpdate
+            return ({}, {})
+
         selected_response = corr_param_resp["response"]
         data = {}
         colors = {}
@@ -382,7 +383,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         else:
             ensembles = None
         if not (ensembles and responses and corr_param_resp["response"] in responses):
-            raise PreventUpdate
+            return {}
         selected_response = corr_param_resp["response"]
         loaded_ensembles = [
             load_ensemble(parent, ensemble_id) for ensemble_id in ensembles
@@ -484,7 +485,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             and corr_param_resp["parameter"] in parameters
             and corr_param_resp["response"] in responses
         ):
-            raise PreventUpdate
+            return ({}, [])
 
         selected_parameter = corr_param_resp["parameter"]
         selected_response = corr_param_resp["response"]
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -62,6 +62,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             ),
         ],
         [
+            Input(parent.uuid("ensemble-selection-store"), "data"),
             Input(parent.uuid("correlation-store-selected-obs"), "data"),
             Input(parent.uuid("correlation-store-selection"), "data"),
             Input(parent.uuid("correlation-metric"), "value"),
@@ -71,10 +72,10 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         [
             State(parent.uuid("parameter-selection-store-param"), "data"),
             State(parent.uuid("parameter-selection-store-resp"), "data"),
-            State(parent.uuid("selected-ensemble-dropdown"), "value"),
         ],
     )
     def update_correlation_plot(
+        ensemble_selection_store: Dict[str, List],
         selected_obs: Dict,
         corr_param_resp: Dict,
         correlation_metric: str,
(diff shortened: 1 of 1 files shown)


valentin-krasontovitsch:
manual testing shows (at least for me) that this fixes [the issue](#326)

valentin-krasontovitsch:
jenkins, retest this please.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -470,6 +465,17 @@ def update_corr_param_resp(
         parent.save_state("active_correlation", corr_param_resp)
         return corr_param_resp
 
+    def get_selected_ensembles_from_store(
```
DanSava, line 468: The function should be "private" I think -> `_get_selected_ensembles_from_store` also it is a good idea to define it before it is used so at the beginning of this file maybe

valentin-krasontovitsch: thank you, of courese, gonna make it private! 
about the beginning or end - this is i guess something we should discuss @hnformentin :sweat_smile:

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -470,6 +465,17 @@ def update_corr_param_resp(
...
+        ensemble_selection_store: Dict[str, List]
+    ) -> Optional[List[str]]:
+        if ensemble_selection_store:
+            ensembles = [
+                selection["value"] for selection in ensemble_selection_store["selected"]
+            ]
+        else:
+            ensembles = None
+        return ensembles
```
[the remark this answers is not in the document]
valentin-krasontovitsch: but then it's harder to print the value for, you know, for the print debugger! ; )

webviz_ert/controllers/response_correlation_controller.py
hnformentin: Regarding vertical ordering, we follow the recommendation in "Clean Code" 📖 : 
<h1>Vertical Ordering</h1>
In general we want function call dependencies to point in the downward direction. That is, a function that is called should be below a function that does the calling*.

As in newspaper articles, we expect the most important concepts to come first, and we expect them to be expressed with the least amount of polluting detail. We expect the low-level details to come last. This allows us to skim source files, getting the gist from the first few functions, without having to immerse ourselves in the details.

* This is the exact opposite of languages like Pascal, C and C++ that enforce functions to be defined, or at least declared, _before_ they are used.

webviz_ert/controllers/response_correlation_controller.py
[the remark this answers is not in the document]
hnformentin: Maybe this would be a solution in the middle :-) 

```
if ensemble_selection_store:
            ensembles = [selection["value"] for selection in ensemble_selection_store["selected"]]
            return ensembles
return None

```

webviz_ert/controllers/response_correlation_controller.py
DanSava: I was not aware of the "clean code" recommendation, can you also provide a link to where this is coming from? 

Anyway, then I think there are a lot of ppl in scout with a C/C++ background, myself included because most of the helper functions are defined at the top of files in most scout projects,  but I guess the bottom of the file is just as fine, not that it really matters if we are talking about a file that is 700+ lines.

webviz_ert/controllers/response_correlation_controller.py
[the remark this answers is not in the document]
DanSava: I think @valentin-krasontovitsch was joking there, but my suggestion was not a hard point, so @valentin-krasontovitsch  is free to do what he wants either keep his original version or use @hnformentin suggestion

DanSava (approved):

webviz_ert/controllers/response_correlation_controller.py
hnformentin: @DanSava it is the book called "Clean Code, a handbook of agile software craftsmanship", by Robert Martin, page 84. We have a printed sample at the office. I can ask in the channel to see what we want as a team.

valentin-krasontovitsch: discussing in the team is a good idea, then we can find a convention we can follow : )

webviz_ert/controllers/response_correlation_controller.py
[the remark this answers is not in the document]
valentin-krasontovitsch: thanks for the suggestions, i can totally make it concise, it was mostly a joke, with a hint of truth in it, as my favorite way of debugging (in python) is indeed just printing variable values...

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: made private!

webviz_ert/controllers/response_correlation_controller.py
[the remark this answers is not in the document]
valentin-krasontovitsch: made concise.

valentin-krasontovitsch:
sorry, this has been approved a while back, i was just wondering if you @DanSava had any comment on me leaving it be as two separate commits?

valentin-krasontovitsch:
i see that dan is on vacation. gonna merge this then

PR #339, opened 2022-07-27
refactor and fix x-index update function

### LINKED ISSUE #337: Response Correlation Viewer: Response overview plot for response with observations doesn't survive reload
**Describe the bug**
The response overview plot does not survive reloads when a response with observations is selected and active.

**To Reproduce**
3. Choose an ensemble, and a response with observations (use tickbox to filter responses!)
4. Refresh / reload the page using your browser
5. The response overview plot, previously drawn, now remains blank

**Expected behavior**
The response overview plot is redrawn, like before the refresh

 - OS: Ubuntu 22.04 "jammy"
 - Browser: Google Chrome 102.0.5005.115

valentin-krasontovitsch:
**Issue**
Closes #337 

**Approach**
We tidy up the edge-cases for populating the correlation x-index.
We also refactor a wee bit, see discussion in the issue.

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -15,6 +15,7 @@ from webviz_ert.models import (
     load_ensemble,
     BarChartPlotModel,
     PlotModel,
+    Response,
 )
 from webviz_ert.controllers.multi_response_controller import (
     _get_observation_plots,
@@ -24,6 +25,8 @@ from webviz_ert import assets
 
 
 def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) -> None:
+    DEFAULT_X_INDEX = 0
+
     @app.callback(
         [
             Output(
@@ -230,13 +233,6 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
                 for obs in response.observations:
                     obs_plots.append(_get_observation_plots(obs.data_df()))
 
-                if corr_xindex[selected_response] == 0 and isinstance(
-                    response_x_axis, (pd.Index, list)
-                ):
-                    corr_xindex[selected_response] = _update_corr_index_dict(
-                        response_x_axis, response.observations[0].data_df()
-                    )
-
         fig = go.Figure()
         for plot in response_plots:
             fig.add_trace(plot.repr)
@@ -247,8 +243,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
 
         fig.update_layout(_layout_figure(x_axis_label))
 
-        default_index = 0
-        x_index = corr_xindex.get(selected_response, default_index)
+        x_index = corr_xindex.get(selected_response, DEFAULT_X_INDEX)
         if isinstance(response_x_axis, pd.Index) and not response_x_axis.empty:
             fig.add_shape(
                 type="line",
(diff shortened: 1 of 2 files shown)


webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -400,6 +400,33 @@ def update_response_parameter_scatterplot(
...
+        Some data shown is a function of the x value represented by that index,
+        for example the correlation scatterplot, as well as the heatmap and the
+        tornado plot.
+
+        This function updates a dictionary with responses as keys, and the
+        current index for each response as values.
+
+        It gets triggered when a user
+        - clicks on the response overview plot, effectively choosing an index
+        - changes the selection of responses
```
hnformentin, line 414: I would add: ", in this case a default value is stored".

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -433,13 +433,15 @@ def update_corr_index(
         ctx = dash.callback_context
         triggered_id = ctx.triggered[0]["prop_id"].split(".")[0]
 
-        if triggered_id == parent.uuid("parameter-selection-store-resp"):
+        if triggered_id == parent.uuid("parameter-selection-store-resp") or (
+            triggered_id == "" and not click_data
```
hnformentin, line 444: Which is the case `triggered_id == ""`? I could not figure it out.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -427,26 +426,46 @@ def update_corr_index(
...
         """
-        if not responses:
+        if ensemble_selection_store:
+            ensembles = [
+                selected_ensemble["value"]
+                for selected_ensemble in ensemble_selection_store["selected"]
+            ]
+        else:
+            ensembles = None
+        if not (responses and ensembles):
```
hnformentin, line 435: If some ensembles and responses were selected before, we are storing the old values. Is this reasoning correct?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -427,26 +426,46 @@ def update_corr_index(
...
+                    response_name, default_index
+                )
         elif click_data:
             corr_xindex[corr_param_resp["response"]] = click_data["points"][0][
                 "pointIndex"
             ]
         return corr_xindex
 
+    def _get_default_x_index(response: Response) -> int:
+        x_axis = response.axis
```
hnformentin, line 460: For me it is confusing that we were using `x_axis` for the x-axis of the response, but `obs_x_axis` for the x-axis of the observation. For symmetry, I would prefer to use the full name for response as well: `response_x_axis`.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -527,16 +546,16 @@ def _define_style_ensemble(index: int, x_axis: pd.Index) -> Dict:
     return style
 
 
-def _update_corr_index_dict(
+def _calculate_first_observation_index(
```
hnformentin, line 548: Would that make sense that the name of the function is `_get_first_observation_index`. It seems would be more uniform with the names of other functions and in this function we are actually not calculating, but getting the index...What do you think?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -241,8 +241,7 @@ def update_response_overview_plot(
 
         fig.update_layout(_layout_figure(x_axis_label))
 
-        default_index = 0
-        x_index = corr_xindex.get(selected_response, default_index)
+        x_index = corr_xindex.get(selected_response)
```
hnformentin, line 244: This `default_index` was added to be explicit about the default selection of this `get` function. It makes it easier to understand that if corr_xindex was empty for a given `selected_response`, `x_index` would be zero. The proposal of this commit is DRYer but let implicit this choice. Do you see a way to be explicit and also DRY?

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: sure, done : )

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: of course, done ^^

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: That is a very good question. At the end of the day, it is up to user preference I suppose - expected behaviour, I feel, could go either way, it gets cleared, or it don't get cleared, I dunno... I chose this behaviour (of remembering) for now. If you know of a preference, or if you think that this is unintuitive / surprising behaviour, I will adjust it!

indeed *right now* the check is too much, we don't need to check for ensembles and responses, no ensembles implies no responses - I thought a little bit about it, and decided to take "defensive programming" approach here, instead of being more technically correct and concise. In case of future changes, where responses might be selected / exist without ensembles being selected, this is not gonna break.

but if you think this is too defensive or the scenario above is too far-fetched, I can adjust it of course : )

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: ya, very good, I thought yesterday "this probably deserves an inline comment, or an explanation in the commit message" : D

that case is you have selected ensembles, and responses, and you refresh the page.
at the moment, the `corr_xindex` store is not of session type - it is not remembered across reloads (I asked Dan about this in an issue, but didn't get a response, he must have overlooked it, and then i forgot until seeing it again yesterday), but ensemble and response selection is remembered.
So you reload, ensembles are chosen, responses are chosen, but if we do not have that check `triggered_id == "" and not click_data`, which is basically the callback firing for the first time on page load, then the `corr_xindex` will not get populated, and you end up without an index.

does that make sense?

if so, what do we do on the team / what would you prefer or suggest: an inline comment, or a descriptive git commit message?

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: indeed, thank you!

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: I do appreciate being explicit, but I'm not sure if I want to be explicit in that spot - if i was reading the code, and wanted to know how that `corr_xindex` gets calculated / set, i would try to find the function manipulating it (in this case, the callback that has it as output), and check there. and indeed, in that callback, `_get_default_x_index` is called, which has as last line `return 0`.

If you feel like we want to be explicit in the plot callback as well, no problem, I can create a variable scoped to the controller, which holds the default index value `0`, and use it in both places (in the `_get_default_x_index` function, and in the response overview plot callback). Then it's easier to see the default value in that context, and it's DRY, so no problem with updating the value.

what do you think?

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: ah i see i actually put it in the commit message of commit 9d3823d26f86dcbd217cdbd721853c69bdf56fdc :sweat_smile: but maybe it is not very clear?

hnformentin: It makes a lot of sense 🎉 
I would suggest to have it documented as code, e.g. having a variable `page_reloaded = (triggered_id == "" and not click_data)`, and then having `if triggered_id == parent.uuid("parameter-selection-store-resp") or page_reloaded`.
I think we are not in the state to trust that we will get much information from a git commit message (I think you are the only one in the squad keeping this good habit so far, we should definitely do it more), and an inline comment would be as ugly as this alternative I suggested, but I prefer to have the programatic comment ;-)  What do you think? Maybe the name `page_reloaded` could be of a better choice?

webviz_ert/controllers/response_correlation_controller.py
hnformentin: Both are good, I was trying to check if my understanding about your intentions was correct, and it was! Keep it going :-)

webviz_ert/controllers/response_correlation_controller.py
hnformentin: I think that is actually a very good suggestion! In this way we are explicit and dry!

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: ah, that's a great solution, thank you ^^
i guess the name could be `first_invocation` ? technically it's not only on reload, i suppose...

valentin-krasontovitsch: used `first_invocation` now.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: done : )

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: great, thanks : )

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -498,16 +550,16 @@ def _define_style_ensemble(index: int, x_axis: pd.Index) -> Dict:
...
 
 
-def _update_corr_index_dict(
+def _get_first_observation_index(
     response_x_axis: Union[pd.Index, list], observation_x_axis: pd.DataFrame
 ) -> int:
     x_axis_default_observation = _get_first_observation_x(observation_x_axis)
     if isinstance(response_x_axis, pd.Index):
-        updated_index = response_x_axis.get_loc(x_axis_default_observation)
+        calculated_index = response_x_axis.get_loc(x_axis_default_observation)
```
hnformentin, line 558: Sorry for late comment, but would not make more sense to be first_observation_index instead of calculated_index?

valentin-krasontovitsch: yes indeed, good catch : ) gonna change and squash all into one commit!

valentin-krasontovitsch: done!

Issue #340, opened 2022-07-28
Add testing for the new observation selector controller


DanSava:
Issue 
https://github.com/equinor/webviz-ert/pull/332 has introduced a new controller with very minimal testing for a small part of the logic in the controller. 

Try to test the controller functionality.

PR #344, opened 2022-07-29
Refactor obs selector

hnformentin:
**Issue**
Resolves partially #340

**Approach**
(1) Extract parts of code into local functions
(2) Try to test those functions
(3) Refactor and loop

### DIFF
--- a/tests/controllers/test_response_correlation_controller.py
+++ b/tests/controllers/test_response_correlation_controller.py
@@ -4,7 +4,6 @@ import pytest
 from pandas._libs.tslibs.timestamps import Timestamp
 
 from webviz_ert.controllers.response_correlation_controller import (
-    _get_first_observation_x,
     _define_style_ensemble,
     _layout_figure,
 )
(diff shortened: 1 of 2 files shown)


hnformentin:
Locally it run, but got an error from mypy. Any idea why running the command locally did not capture this problem?

```
> Run mypy --config-file mypy.ini --package webviz_ert
> webviz_ert/controllers/response_correlation_controller.py:60[7](https://github.com/equinor/webviz-ert/runs/7627551261?check_suite_focus=true#step:5:8): error: "tuple" is not subscriptable, use "typing.Tuple" instead  [misc]
> Found 1 error in 1 file (checked 39 source files)
> Error: Process completed with exit code 1.
```

valentin-krasontovitsch:
> Any idea why running the command locally did not capture this problem?

only thing i can think of - different versions of mypy?

hnformentin:
Closed before review and merge since #353 reverted the content that was refactored.

PR #353, opened 2022-08-02
Revert "New obs selector controller"

valentin-krasontovitsch:
This reverts commit df14a76898b1830ab961028cdaf32255d0830f36.

We have decided that the merged feature was not quite ready yet and
perhaps introduced to many bugs. We shall instead try to introduce the
feature more gradually.


hnformentin (approved):
We even tested the frontend ;-)

PR #354, opened 2022-08-03
Refactor `update_corr_resp_param`

### LINKED ISSUE #336: Response Correlation Viewer: Refactor controller function `update_corr_param_resp`
Trying to understand the usage of the output `correlation-store-selection` of said function in other callbacks, we observe that the logic of the function is difficult to read (at least for me) and the state handling is not necessarily intuitive / as expected (again, at least for me).

I think some refactoring might help with clearing up the way state is handled, both logically and from a presentational point of view.

valentin-krasontovitsch:
**Issue**
Resolves #336

**Approach**

- refactor the logic of data handling in the callback function
- add a doc string describing the expected behaviour of the callback function
- rename the callback function, the store, and variables derived from said store

**Notes for reviewer**
Please have a critical look at the logic refactoring of the callback - I don't know if I actually improved it (much).

The checks in various other callbacks of the form

```
active_resp_param["parameter"] in parameters
```

are left in place, even though they were what originally made me stumble upon this function - they are strictly speaking not necessary, as an unselected parameter (`parameters` contains the selected parameters) will never be active (and deselection of an active parameter shifts active state to another selected parameter). Nevertheless I decided not to remove them, to continue with the idea of defensive programming.

Now it really makes me wonder though if one should leave it in, because it has the tendency to confuse people (or at least me!) - you just have to wonder, is this possible state?! but perhaps thanks to the docs it's clear what is possible state and what is not. also, the defensive approach could be argued as isolation of callback between one another - one callback should not rely exactly on the implementation of another callback? I don't know...

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -41,7 +41,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         [
             Input(parent.uuid("ensemble-selection-store"), "data"),
             Input(parent.uuid("correlation-store-xindex"), "data"),
-            Input(parent.uuid("correlation-store-selection"), "data"),
+            Input(parent.uuid("correlation-store-active-resp-param"), "data"),
             Input(parent.uuid("correlation-metric"), "value"),
             Input(parent.uuid("sort-parameters"), "on"),
             Input(parent.uuid("hide-hover"), "on"),
@@ -54,7 +54,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
     def update_correlation_plot(
         ensemble_selection_store: Dict[str, List],
         corr_xindex: Dict,
-        corr_param_resp: Dict,
+        active_resp_param: Dict,
         correlation_metric: str,
         sort_parameters: bool,
         hide_hover: bool,
@@ -66,11 +66,11 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             ensembles
             and parameters
             and responses
-            and corr_param_resp["response"] in responses
+            and active_resp_param["response"] in responses
         ):
             return ({}, {})
 
-        selected_response = corr_param_resp["response"]
+        active_response = active_resp_param["response"]
         data = {}
         colors = {}
         heatmaps = []
@@ -92,10 +92,10 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
 
             corrdf = parameter_df.corr(method=correlation_metric)
             corrdf = corrdf.drop(valid_responses, axis=0).fillna(0)
-            if selected_response not in valid_responses:
+            if active_response not in valid_responses:
                 continue
             if sort_parameters:
-                corrdf, df_index = sort_dataframe(corrdf, df_index, selected_response)
+                corrdf, df_index = sort_dataframe(corrdf, df_index, active_response)
             # create heatmap
             data_heatmap = {
                 "type": "heatmap",
@@ -112,8 +112,8 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             }
             heatmaps.append(data_heatmap)
             ens_key = repr(ensemble)
-            data[ens_key] = corrdf[selected_response]
-            data[ens_key].index.name = selected_response
+            data[ens_key] = corrdf[active_response]
+            data[ens_key].index.name = active_response
             colors[ens_key] = assets.get_color(index=index)
         if data and heatmaps:
             correlation_plot = BarChartPlotModel(data, colors)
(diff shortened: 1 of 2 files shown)


hnformentin:
@valentin-krasontovitsch Nice PR, though it would be nice if you could rebase on main branch before putting for review, in this way we don't get any confusion on what is already merged and what is new changes.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -490,6 +490,31 @@ def update_corr_param_resp(
...
         corr_param_resp: Dict,
     ) -> Optional[Dict]:
+        """
+        update_corr_param_resp is a callback function responsible for handling
+        which response and parameter are active.
+
+        The correlation plots display correlation between a given response and
+        paramater (at an index, see `corr_xindex`) for a chosen ensemble. Hence
+        the need to keep track of which of potentially many selected responses
+        and parameters, respectively, is currently active.
```
hnformentin, line 501: "are" currently active

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -490,6 +490,31 @@ def update_corr_param_resp(
...
+
+        It should be triggered upon
+        - modifying the selection of parameters or responses
+        - clicking on the heatmap, effectively choosing an active response and
+          parameter
```
hnformentin, line 506: choosing "a combination of" an active response and parameter. Suggestion because technically we need to choose them simultaneously in the the heatmap.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -490,6 +490,31 @@ def update_corr_param_resp(
...
+
+        To avoid saying "response or parameter", we refer to either
+        collectively by "variable".
+
+        When the first variable is selected, they become active.
+        Further selection of variables does not modify the active variable.
+        Deselection of the active variable makes the first selected variable
+        active, otherwise deselection does not affect the active state.
+
+        Clicking on the heatmap makes the resoponse and parameter active that
```
hnformentin, line 516: makes the "combinations of " ... if you agreed with the previous suggestion.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -489,7 +489,7 @@ def update_corr_param_resp(
         responses: List[str],
         parameters: List[str],
         corr_param_resp: Dict,
-    ) -> Optional[Dict]:
+    ) -> Dict:
```
hnformentin, line 493: Do you understand why before we had Optional?

valentin-krasontovitsch:
> it would be nice if you could rebase on main branch before putting for review

sorry about that, see it now! will try to avoid it in the future

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: thanks : ) i was puzzling over that one when i wrote it, i really wish i could express myself in a simpler way...

valentin-krasontovitsch: done

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: very good, thank you! and done.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: sure ^^ and done.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: i tried to google it in the context of dash callbacks, and found nothing.

in general i guess it's for when you want to have a `return None` in your code sometimes. but we don't haz that. I thought it was connected to the `raise PreventUpdate` that we had here or there, but I suppose in that case we are raising an exception, and not returning control flow in the usual manner, so the typing of the return shouldn't apply!? i didn't get too deep into the latter point.

in any way, i think it's just not necessary, as we always have a return value, and it's type is always `Dict`. And maybe we also want there to always be a dictionary returned?

does that make sense?

hnformentin: yes, ok! In any case, if it was necessary we will find out very soon :D

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -517,27 +517,25 @@ def update_corr_param_resp(
...
+            ):
+                corr_param_resp["response"] = responses[0]
+        elif triggered_id == parent.uuid("parameter-selection-store-param"):
+            if not parameters:
+                corr_param_resp["parameter"] = None
+            elif (
+                not corr_param_resp["parameter"]
+                or not corr_param_resp["parameter"] in parameters
+            ):
+                corr_param_resp["parameter"] = parameters[0]
```
hnformentin: It is much easier to understand! I just still don't feel comfortable that we have `corr` appended in so many variables, and I don't know exactly if they only refer to `correlation viewer` or something else. If you have an insight on that, please share with me!

webviz_ert/plugins/_response_correlation.py
```diff
@@ -33,31 +33,31 @@ def tour_steps(self) -> List[Dict[str, str]]:
             {
                 "id": self.uuid("response-overview"),
                 "content": (
-                    "Visualization of the currently selected response "
+                    "Visualization of the currently active response "
                     "where by clicking we can select a new index / timestep for correlation analysis."
```
hnformentin, line 37: where by clicking we can select "and activate" ...

webviz_ert/controllers/response_correlation_controller.py
hnformentin: @valentin-krasontovitsch

hnformentin (approved):
LGTM. We tested together locally.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: so one hint I see is the naming of the store - it was `correlation-store-selection`, which to me indicates that this is just referring to the correlation viewer. I don't know, this is just an interpretation.
i guess we can merge for now, and if somebody complains about the naming later, we can change it back : )

webviz_ert/plugins/_response_correlation.py
valentin-krasontovitsch: good catch, thanks, although i was gonna wait with doing renaming on / around the index, but sure!

valentin-krasontovitsch: done

PR #356, opened 2022-08-03
Label the scatterplot 

### LINKED ISSUE #355: Response Correlation Viewer: Display which response and parameter are plotted in the scatterplot
**Is your feature request related to a problem? Please describe.**
Looking at the scatter plot right now, it's a bit hard to parse. There are no labels besides the scales of the axes.

**Describe the solution you'd like**
We woud like to be able to see
- correlation between which response and parameter is displayed, at which index

**Describe alternatives you've considered**
Just looking at the bottom of the page, to the info text, is a bit far from the plot and seems to lead to bad user experience.

valentin-krasontovitsch:
**Issue**
Closes #355

**Approach**
We add titles, and fix hover texts.

In a silly way, some options available in the plotly API did not have any effect when I tried to use them, for instance `x_title` should add a title at the bottom of the scatterplot/histogram combination, but, as I said, no effect : / so it's on the side!

## Pre review checklist

I did this on top of the work of #354, so we should wait for that one to be merged! making this a draft for now.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -4,7 +4,7 @@ pytest-mock
 pylint
 black
 scipy
-dash[testing]
+dash[testing]>=2.5.1
 selenium
 types-requests
 types-pkg_resources
--- a/tests/controllers/test_response_correlation_controller.py
+++ b/tests/controllers/test_response_correlation_controller.py
@@ -5,10 +5,11 @@ from webviz_ert.controllers.response_correlation_controller import (
     _define_style_ensemble,
     _get_first_observation_index,
     _layout_figure,
+    _format_index_value,
 )
 
 from webviz_ert.controllers.response_correlation_controller import (
-    sort_dataframe,
+    _sort_dataframe,
 )
 
 
@@ -20,7 +21,7 @@ def test_sort_dataframe():
     data[other_key] = [0.3, 0.8, -0.1]
     dataframe = pd.DataFrame(data=data)
     index = None
-    sorted_dataframe, _ = sort_dataframe(dataframe, index, other_key)
+    sorted_dataframe, _ = _sort_dataframe(dataframe, index, other_key)
     assert list(sorted_dataframe[other_key]) == sorted(data[other_key])
     assert list(sorted_dataframe[one_key]) == [0.2, -0.4, 0.6]
 
@@ -114,3 +115,19 @@ def test_get_first_observation_x_invalid():
     df_observation = pd.DataFrame([int(1)], columns=["x_axis"])
     with pytest.raises(ValueError, match="invalid obs_data type"):
         _get_first_observation_x(df_observation)
+
+
+@pytest.mark.parametrize(
+    "raw_value,expected_formatted_value",
+    [
+        ("2022-08-05 14:25:00", "2022-08-05"),
+        ("2022-09-21 14:25:00", "2022-09-21"),
+        (14, "14"),
+        ("213", "213"),
+        ("SPAM", "SPAM"),
+    ],
+    ids=["date1", "date2", "numeric", "numeric-as-string", "silly-string"],
+)
+def test_format_index_value(raw_value: str, expected_formatted_value: str):
+    axis = pd.Index(data=[raw_value])
+    assert _format_index_value(axis, 0).__str__() == expected_formatted_value
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -1,6 +1,8 @@
 import pandas as pd
 import plotly.graph_objects as go
 import dash
+import datetime
+import numpy as np
 
 from typing import List, Optional, Dict, Tuple, Any, Union
 from copy import deepcopy
@@ -95,7 +97,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             if active_response not in valid_responses:
                 continue
             if sort_parameters:
-                corrdf, df_index = sort_dataframe(corrdf, df_index, active_response)
+                corrdf, df_index = _sort_dataframe(corrdf, df_index, active_response)
             # create heatmap
             data_heatmap = {
                 "type": "heatmap",
@@ -322,51 +324,44 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
                         PlotModel(
                             x_axis=x_data.values.flatten(),
                             y_axis=y_data.values.flatten(),
-                            text="Mean",
-                            name=f"{repr(ensemble)}: {active_response}x{active_parameter}@{response.axis[x_index]}",
+                            name=(f"{ensemble.name}"),
+                            hoverlabel={"namelength": -1},
                             **style,
                         )
                     ]
                     _resp_plots[str(ensemble)] = x_data.values.flatten()
                     _param_plots[str(ensemble)] = y_data.values.flatten()
 
-        fig = make_subplots(
-            rows=4,
-            cols=2,
-            specs=[
-                [{"colspan": 2, "rowspan": 3}, None],
-                [None, None],
-                [None, None],
-                [{"rowspan": 1}, {"rowspan": 1}],
-            ],
-        )
-        for plot in _plots:
-            fig.add_trace(plot.repr, 1, 1)
+        n_rows = 10
+        histogram_row_span = 2
+        spacer_row_span = 1
+        row_histograms = n_rows - (histogram_row_span // 2)
 
-        for key in _param_plots:
-            fig.add_trace(
-                {
-                    "type": "histogram",
-                    "x": _param_plots[key],
-                    "showlegend": False,
-                    "marker_color": _colors[key],
-                },
-                4,
-                1,
-            )
-        for key in _resp_plots:
-            fig.add_trace(
-                {
-                    "type": "histogram",
-                    "x": _resp_plots[key],
-                    "showlegend": False,
-                    "marker_color": _colors[key],
-                },
-                4,
-                2,
-            )
-        fig.update_layout(assets.ERTSTYLE["figure"]["layout"])
-        fig.update_layout(showlegend=False)
+        specs = _create_scatterplot_specs(
+            n_rows, histogram_row_span, spacer_row_span, row_histograms
+        )
+        fig = make_subplots(
+            rows=n_rows,
+            cols=2,
+            specs=specs,
+        )
+
+        formatted_x_value = _format_index_value(response.axis, x_index)
+        _set_scatterplot_axes_titles(
+            fig, row_histograms, active_response, active_parameter, formatted_x_value
+        )
+
+        _style_scatterplot(fig)
+
+        _add_plots_to_multiplot(
+            fig,
+            _plots,
+            _param_plots,
+            _resp_plots,
+            _colors,
+            row_histograms,
+        )
+
         final_text = []
         for response_name in responses:
             x_axis = ensemble.responses[response_name].axis
@@ -552,7 +547,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         return None
 
 
-def sort_dataframe(
+def _sort_dataframe(
     dataframe: pd.core.frame.DataFrame,
     index: Optional[pd.core.indexes.base.Index],
     sort_by_column: str,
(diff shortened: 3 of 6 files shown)


valentin-krasontovitsch:
sorry forgot to add screenshots, here we go - before the changes:

scatterplot-labels-before-mini
after the changes:

scatterplot-labels-after-mini

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -551,6 +568,16 @@ def _get_selected_ensembles_from_store(
...
+        if isinstance(raw_value, str):
+            try:
+                datetime_value = pd.to_datetime(raw_value)
+                return datetime_value.date()
+            except (pd.errors.ParserError):
+                pass
+        return raw_value
+
 
 def sort_dataframe(
```
hnformentin, line 582: Should it be internal `_`?

valentin-krasontovitsch: the `sort_dataframe` function? outside the scope of this PR, but following the campsite rule, i guess we can make it internal. i tried to make it reusable, but the part where you check if the index exists, that's probably very particular to our case.

valentin-krasontovitsch:
tooltips have also been improved - before:

scatterplot-labels-tooltips-before

after:

scatterplot-labels-tooltips-after

disclaimer: images were photoshopped to overlay two tooltips in one image :innocent:

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -551,6 +568,16 @@ def _get_selected_ensembles_from_store(
             ]
         return None
 
+    def _format_index_value(axis: pd.Index, index: int) -> Union[int, datetime.date]:
```
hnformentin, line 571: I think that it would be more consistent with the other functions if we had them all after `controller`. What do you think?

valentin-krasontovitsch: yes sorry, of course! done.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: done

hnformentin:
Testing the functionality...a surprising behaviour for me was that the name of the response and parameter was not in the axis, but in the title of the scatterplot. As a user I would expect to have these names in the positions that I tried to represent in the image below :D 
Screenshot 2022-08-04 at 14 44 32

Do you think that my surprise adds something, or are you surprised with my surprise?

hnformentin:
But in any case, the proposal is already better than not having any indication of what was parameter and what was response 🎉

tests/controllers/test_response_correlation_controller.py
```diff
@@ -8,19 +8,19 @@
 )
 
 from webviz_ert.controllers.response_correlation_controller import (
-    sort_dataframe,
+    _sort_dataframe,
 )
 
 
-def test_sort_dataframe():
+def test__sort_dataframe():
```
hnformentin, line 15: Different than the tests for the other internal functions, where I (not sure if we) used only one `_`. Do you think we should change all of them to be consistent though the test file?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -322,14 +323,16 @@ def update_response_parameter_scatterplot(
                         PlotModel(
                             x_axis=x_data.values.flatten(),
                             y_axis=y_data.values.flatten(),
-                            text="Mean",
-                            name=f"{repr(ensemble)}: {active_response}x{active_parameter}@{response.axis[x_index]}",
+                            text="Mean",  # TODO ask about this!
```
hnformentin, line 326: `# TODO ask about this!` what is this comment about?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -610,3 +627,14 @@ def _get_first_observation_x(obs_data: pd.DataFrame) -> Union[int, str]:
...
+
+def _format_index_value(axis: pd.Index, index: int) -> Union[int, datetime.date]:
+    raw_value = axis[index]
+    if isinstance(raw_value, str):
+        try:
+            datetime_value = pd.to_datetime(raw_value)
+            return datetime_value.date()
+        except (pd.errors.ParserError):
+            pass
+    return raw_value
```
hnformentin, line 640: This function could be tested. It would be good to document which states we expect to have a proper value and which states the exception is to be raised.

hnformentin:
> **Issue** Closes #355
> 
> **Approach** We add titles, and fix hover texts.
> 
> In a silly way, some options available in the plotly API did not have any effect when I tried to use them, for instance `x_title` should add a title at the bottom of the scatterplot/histogram combination, but, as I said, no effect : / so it's on the side!
> 
> ## Pre review checklist
> * [x]  Added appropriate labels
> 
> I did this on top of the work of #354, so we should wait for that one to be merged! making this a draft for now.

I was able to have proper titles on the axis of the scatterplot. Just added:
`
fig.update_layout({"xaxis": {"title": {"text": f"{active_response} @ {formatted_x_value}"}}})
fig.update_layout({"yaxis": {"title": {"text": f"{active_parameter}"}}})
`
after
`
fig.update_layout(assets.ERTSTYLE["figure"]["layout"])
fig.update_layout(showlegend=False)
`

and got 

Screenshot 2022-08-05 at 09 03 38

Now for the histogram it is another problem :-) I am trying to see where the layout properties of the histograms are, but so far did not find it. Any idea?

tests/controllers/test_response_correlation_controller.py
valentin-krasontovitsch: ah, this was just a result of blind replacing / renaming. gonna fix it in this instance, don't think we need a double underscore, it's just a name after all : )

and done.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: XD that's hilarious, maybe they put it to have something? my comment is indeed for finding out why this is called mean, and i also thought that it didn't make sense, but since i don't know much about the domain or statistics, i thought i'd ask ^^

ok i figured out that no plot was drawn because the plot model wants to get a name. extending the plot model to accept no given name, we fix the no plot and circumvent having empty space / an empty line in the tooltip. yeahy : )

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: thanks for pointing that out - I wrote a parametrized test, and found an error (or not considered case) in the code, so that was nice ^^

have a look and see if it's alright please : )

valentin-krasontovitsch:
i am completely confused about the roles of layout and non-layout properties, but it seems like I found a way of settings axes titles (found in [plotly examples about subplots](https://plotly.com/python/subplots/) under **Customizing Subplot Axes**)

pushed the changes, observe that there's some potentially silly half-DRY business going on with storing the number of rows in a variable.

Also please observe that I have made space between the subplots, which makes the scatterplot  a bit smaller.

here's the changes isolated

new-titles-mini

and here it is in context

new-titles-in-context

I might try to play with it, to decrease the size of the spacing a bit more!

but for now - what do you think, Helena?

tests/controllers/test_response_correlation_controller.py
```diff
@@ -114,3 +115,18 @@ def test_get_first_observation_x_invalid():
...
+@pytest.mark.parametrize(
+    "raw_value,expected_formatted_value",
+    [
+        ("2022-08-05 14:25:00", "2022-08-05"),
+        ("2022-09-21 14:25:00", "2022-09-21"),
+        (14, "14"),
+        ("213", "213"),
+        ("SPAM", "SPAM"),
+    ],
+)
```
hnformentin, line 129: Very nice, I would just suggest the order of the tests and rename one of them. In this way, the test cases are clearly expressing what is written in the string of the function:
```
@pytest.mark.parametrize(
    "raw_value,expected_formatted_value",
    [
        ("2022-08-05 14:25:00", "2022-08-05"),
        ("2022-09-21 14:25:00", "2022-09-21"),
        ("PARSING_FAILS_FROM_NOW", "PARSING_FAILS_FROM_NOW"),
        (14, "14"),
        ("213", "213"),
    ],
)
```

hnformentin:
> 

That is really nice @valentin-krasontovitsch !
You managed to add all these labels. I am curious about how you implemented it 🤓 

Ultimately, it would be better to ask for user feedback directly, nevertheless, some considerations that I would like to share.

If I were a user, I would be happy to have the labels on the axes of the plots, not necessarily the titles of them. Specially if I needed to choose between having larger plots or their titles. This seems to be a good point to ask feedback, and since we already there with the implementation of the titles, it would be cool to let the code related to the titles/size flexible (if possible) and ask for feedback later when they are back.

It is really nice to have quick access to all the ensembles selected 👏 Nevertheless, if I were a user, I would prefer to analyse the scatterplot and histograms of each ensemble in a plot separately. This is because when we look for these kind of plots, we try to capture trends within the ensemble data and outliers. 

Some context about data analysis, if you are interested:
- With the scatterplot we can observe trends that diverge from linear correlation (since the correlation coefficient that we calculate represents only linear correlation...imagine that a synodal correlation would correspond to a zero linear correlation, displayed in the heat map) and outliers, which are points that deform the calculated correlation, and maybe not representative of the phenomena been modelled or even lead to investigate a new phenomena that we were not aware about.
- With the histograms we can observe trends that diverge from normal distribution (since many statistical methods are based on this assumption) and outliers.

Having multiple ensembles plotted at the same time prevent me to see the trends and outliers within an ensemble. At the same time, having the possibility of displaying  different ensembles is a nice feature to analyse changes between two or several ensembles. But we can do this by clicking in the heat map at same parameters x response combination, but different ensembles. If this becomes too difficult because of a larger number of parameters and responses (and consequently very small rectangles displayed on the heat map), I would prefer to selector buttons with which I could choose dynamically the ensemble to be plotted.

Does any of this make sense?

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -610,3 +628,32 @@ def _get_first_observation_x(obs_data: pd.DataFrame) -> Union[int, str]:
...
+    active_parameter: str,
+    x_value: Union[datetime.date, np.number],
+    n_rows: int,
+) -> None:
+    fig.update_layout(assets.ERTSTYLE["figure"]["layout"])
+    fig.update_xaxes(title_text=f"{active_response} @ {x_value}", row=1, col=1)
+    fig.update_yaxes(title_text=f"{active_parameter}", row=1, col=1)
+    fig.update_xaxes(title_text=f"{active_parameter}", row=n_rows, col=1)
+    fig.update_xaxes(title_text=f"{active_response}", row=n_rows, col=2)
+    fig.update_layout(showlegend=False)
```
hnformentin, line 659: In other parts of the code, we have updated the layout via a dictionary and not passing the `fig` element, e.g. `_define_style_ensemble` or `_layout_figure`. This was chosen to have simpler variables been passed and consequently easier to test if we want so. What would you think about maintaining a similar pattern for layout functions?

valentin-krasontovitsch: are you referring just to the last line, 659, setting visibility of the legend? or also to the `update_?axes` calls?

if you mean the latter, I can try to see if this can be replicated by passing a dict to `update_layout`

tests/controllers/test_response_correlation_controller.py
valentin-krasontovitsch: very good suggestion - it seems like we can actually add descriptive names to our test cases, cf. [the pytest parametrize docs on test IDs](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids)

i'm gonna try to have a go at that!

valentin-krasontovitsch: I want to insist a little bit on keeping the `SPAM` as a silly string, as it seems to be [the pythonic way](https://docs.python.org/3.10/library/pprint.html) (see first code block) - but it's not a hard insist : )

valentin-krasontovitsch: considering your suggestion of `PARSINGS_FAILS_FROM_NOW`, I would like to avoid adding that information ("parsing fails") to the test cases - i think it's an implementation details of the function. what the test wants to know, is: if I put in something that is not a date, will I get out the same thing, possibly converted to a string if it was a number.

does that sound OK? pushed the updated version with IDs

valentin-krasontovitsch:
> That is really nice @valentin-krasontovitsch ! 

Thank you ^^

> You managed to add all these labels. I am curious about how you implemented it nerd_face

i know, right? :grin:  in some other docs / example i found the methods `update_xaxes` and `update_yaxes` used on the figure object in a multiplot context to achieve exactly what we wanted! just needed to google moar... a link is in [a comment above on this PR](https://github.com/equinor/webviz-ert/pull/356#issuecomment-1206513766)

> If I were a user, I would be happy to have the labels on the axes of the plots, not necessarily the titles of them. 

but the labels are on the axes now, aren't they? :thinking: see the [screenshots above, in this comment](https://github.com/equinor/webviz-ert/pull/356#issuecomment-1206513766)
i just added actual titles, like `Histogram Response`, but I can remove those if you think it's too much / superfluous information : ) very flexible there, no problem at all!

> It is really nice to have quick access to all the ensembles selected clap 

I'm sorry, but I cannot take credit for that - the multiple ensembles setup was already done before I started working on the labels (and are also not in the scope of this issue / PR).

If you think the current layout / display with multiple ensembles is not good, we could probably make an issue about it, and discuss with whoever came up with the design in the first place. 

> Some context about data analysis, if you are interested:
> 
> ...
> 
> Does any of this make sense?

Thank you, it definitely makes a lot of sense, and is very helpful ^^

When you explain it like that, I have to say that I agree and don't know why they designed it the way it is at the moment! Based on the git history, this seems to have been introduced in the original commit 194838f21ec81d82e9ab398e8ad4011c21d96540 that created the whole plugin, and @xjules was the author : )

Julius, do you have any comments? ^^

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: i tried to take a look. following assumptions were made: using the `update_layout` function, we need to pass a dict (cf. [docs on `update_layout`)[https://plotly.com/python-api-reference/generated/plotly.graph_objects.Figure.html?highlight=update_layout#plotly.graph_objects.Figure.update_layout]. that dict should be of type [`plotly.graph_objects.layout`](https://plotly.github.io/plotly.py-docs/generated/plotly.graph_objects.Layout.html#plotly.graph_objects.Layout)

Now, that one has a property `Grid` where one can supply `xaxes`, but that one should be a list of strings matching some regex (`^x([2-9]|[1-9][0-9]+)?( domain)?$`), which among other things means "starts with x" - this does not seem to be what we need. also, the docs for that property mention

> Used with yaxes when the x and y axes are shared across columns and rows

so again this doesn't seem to be what we need, as we don't want to have axes shared across columns and rows

trying to set `xaxes` directly  in the layout dict (without `grid`) gives an error message.

it looks like the `update_?axes` methods cannot be replaced with a layout dict : /

tests/controllers/test_response_correlation_controller.py
hnformentin: That is fine!

webviz_ert/controllers/response_correlation_controller.py
hnformentin: I understand. Thank you for verifying that it is not possible and writing it out!

hnformentin:
> i know, right? 😁 in some other docs / example i found the methods `update_xaxes` and `update_yaxes` used on the figure object in a multiplot context to achieve exactly what we wanted! just needed to google moar... a link is in [a comment above on this PR](https://github.com/equinor/webviz-ert/pull/356#issuecomment-1206513766)

Thanks for sharing it!

hnformentin:
> but the labels are on the axes now, aren't they? 🤔 see the [screenshots above, in this comment](https://github.com/equinor/webviz-ert/pull/356#issuecomment-1206513766)
> i just added actual titles, like `Histogram Response`, but I can remove those if you think it's too much / superfluous information : ) very flexible there, no problem at all!

Yes, I thought was too much to have these titles because I would see better if there was more space to the plots. But it is up to you!

hnformentin:
> I'm sorry, but I cannot take credit for that - the multiple ensembles setup was already done before I started working on the labels (and are also not in the scope of this issue / PR).
> 
> If you think the current layout / display with multiple ensembles is not good, we could probably make an issue about it, and discuss with whoever came up with the design in the first place.
> 
> > Some context about data analysis, if you are interested:
> > ...
> > Does any of this make sense?
> 
> Thank you, it definitely makes a lot of sense, and is very helpful ^^
> 
> When you explain it like that, I have to say that I agree and don't know why they designed it the way it is at the moment! Based on the git history, this seems to have been introduced in the original commit [194838f](https://github.com/equinor/webviz-ert/commit/194838f21ec81d82e9ab398e8ad4011c21d96540) that created the whole plugin, and @xjules was the author : )
> 
> Julius, do you have any comments? ^^

Indeed I did not notice we had that before!!! Thank you for doing the search.
And then I am curious why we have that since the beginning...

hnformentin (approved):

valentin-krasontovitsch:
> Yes, I thought was too much to have these titles because I would see better if there was more space to the plots. But it is up to you!

right, sorry, I said I would try to make the plots bigger, but forgot :sweat_smile: 
now I tried it out - just doubling the number of rows, and scaling the plots accordingly, effectively shrinking the empty row used as space for the axes labels and titles.

here's the result, once with titles, once without

squeezed-with-titles

squeezed-without-titles

~~i would say the space is needed in any case for the axes labels~~ looking at the screenshots, its apparent that the plots become bigger without titles,so I'm gonna drop them for now, and other plots don't have no titles either, so that will be more consistent, too.

i have on my todo to make an issue about adding more visual hints in the response correlation viewer - titles for all plots could be part of that : )

valentin-krasontovitsch:
.... and i forgot to adjust one index, which led to the disappearing of the x-axes labels of the bottom plots - gonna fix that.

webviz_ert/controllers/response_correlation_controller.py
valentin-krasontovitsch: no problem! : )

valentin-krasontovitsch:
OK, I've refactored a bit to make the numbers games I'm playing with rows in the multiplot a bit more DRY, and pulled stuff out into helper function.

Will you please take a look at the last commit, @hnformentin?

Latest screenshot:

final-scatterplot

valentin-krasontovitsch:
Re: multiple ensembles - according to @xjules the current layout was requested from the previous product owner. he agrees that it feels cluttered at the moment. he advises further to wait for @oysteoh who might have comments on this particular aspect from actual users.

valentin-krasontovitsch:
had to quickly fix a flaky test, that's that last commit from just now, sorry. hope it's OK to take it in this PR as it's a quick fix, using the known solution : )

valentin-krasontovitsch:
and another commit, this time for fixing the dash version so that the test rewrite for fixing the flakiness works in jenkins

test_requirements.txt
```diff
@@ -4,7 +4,7 @@ pytest-mock
 pylint
 black
 scipy
-dash[testing]
+dash[testing]==2.5.1
```
hnformentin, line 7: Why do we need to pin?

valentin-krasontovitsch: for some reason (i'm not sure why, i don't understand the inner working of pip orhow it resolves dependencies) the jenkins run was installing dash 2.4, and that one didn't have the method `wait_for_class_to_equal` that i needed for testing... so i just bumped it to a higher version.

but maybe, is there a notation like `>=2.5` that can be used? gonna google that real quick...

valentin-krasontovitsch: fixed, it now says `>=2.5.1`

valentin-krasontovitsch:
thank you : )

the currently latest screenshot displays the latest state - the test and version stuff didn't affect the layout, i hope ^^

so i'm gonna rebase, push, wait for tests to pass again, and then merge

PR #358, opened 2022-08-05
Address flakiness 297

### LINKED ISSUE #297: Adress flakiness in test_axes_labels
**Describe the bug**
There is a flakiness associated with test_axes_labels. It fails randonly, particularly in Github Actions.
Check https://ci.equinor.com/scout/job/komodo_testing/job/webviz-ert-pr/334/console .

### LINKED ISSUE #360: flakiness in tests/views/test_response_correlation.py::test_show_respo_with_obs
(partial) jenkins output:

```
=================================== FAILURES ===================================
___________________________ test_show_respo_with_obs ___________________________

    def test_show_respo_with_obs(mock_data, dash_duo):
        """Test response observation filter works as expected"""

        ...
    
        expected_def_responses = ["FGPT", "FOPR", "SNAKE_OIL_GPR_DIFF", "WOPR:OP1"]
        assert default_responses == expected_def_responses
    
        expected_obs_responses = ["FOPR", "WOPR:OP1"]
        obs_radio_btn = dash_duo.find_element(
            "#" + plugin.uuid("response-observations-check")
        )
        obs_radio_btn.click()
        obs_responses = get_options(dash_duo, "#" + resp_selector_id)
>       assert obs_responses == expected_obs_responses
E       AssertionError: assert ['FGPT', 'FOP...', 'WOPR:OP1'] == ['FOPR', 'WOPR:OP1']
E         At index 0 diff: 'FGPT' != 'FOPR'
E         Left contains 2 more items, first extra item: 'SNAKE_OIL_GPR_DIFF'
E         Full diff:
E         - ['FOPR', 'WOPR:OP1']
E         + ['FGPT', 'FOPR', 'SNAKE_OIL_GPR_DIFF', 'WOPR:OP1']

tests/views/test_response_correlation.py:133: AssertionError
```

looks to me like the button press was not processed.

valentin-krasontovitsch:
**Issue**
Closes #297
Closes #360

**Approach**
try to reproduce the flakiness here (most recently observed in komodo bleeding build, and led to failed build!), and fix it, verify by triggering komodo tests / build, or waiting til next day (automated bleeding build overnight)

## Pre review checklist

### DIFF
--- a/tests/views/test_response_correlation.py
+++ b/tests/views/test_response_correlation.py
@@ -76,21 +76,20 @@ def test_axes_labels(mock_data, dash_duo):
         response_selector_id = plugin.uuid("parameter-selector-multi-resp")
         select_by_name(dash_duo, f"#{response_selector_id}", wanted_response)
 
-        # wait a bit for the graph to be drawn
-        wait_a_bit(dash_duo)
-
         plot_id = plugin.uuid("response-overview")
 
         # check that y axis label spells out "Value"
-        y_axis_title = dash_duo.find_element(f"#{plot_id} text.ytitle")
-        assert y_axis_title.text == "Value"
+        dash_duo.wait_for_text_to_equal(f"#{plot_id} text.ytitle", "Value", timeout=5)
 
         # check that one has date, the other has index as x axis label
-        x_axis_title = dash_duo.find_element(f"#{plot_id} text.xtitle")
         if wanted_response == "FOPR":
-            assert x_axis_title.text == "Date"
+            dash_duo.wait_for_text_to_equal(
+                f"#{plot_id} text.xtitle", "Date", timeout=5
+            )
         else:
-            assert x_axis_title.text == "Index"
+            dash_duo.wait_for_text_to_equal(
+                f"#{plot_id} text.xtitle", "Index", timeout=5
+            )
 
         # clear response selection so next selection can be displayed
         response_deactivator_id = plugin.uuid("parameter-deactivator-resp")
--- a/tests/views/test_response_correlation.py
+++ b/tests/views/test_response_correlation.py
@@ -5,7 +5,6 @@ from tests.conftest import (
     select_by_name,
     select_ensemble,
     wait_a_bit,
-    get_options,
 )
 
 
@@ -117,16 +116,20 @@ def test_show_respo_with_obs(mock_data, dash_duo):
 
     select_ensemble(dash_duo, plugin, "default3")
 
-    resp_selector_id = plugin.uuid("parameter-selector-multi-resp")
-    default_responses = get_options(dash_duo, "#" + resp_selector_id)
+    expected_default_responses = ["FGPT", "FOPR", "SNAKE_OIL_GPR_DIFF", "WOPR:OP1"]
+    expected_responses_with_observations = ["FOPR", "WOPR:OP1"]
 
-    expected_def_responses = ["FGPT", "FOPR", "SNAKE_OIL_GPR_DIFF", "WOPR:OP1"]
-    assert default_responses == expected_def_responses
+    response_selector_id = "#" + plugin.uuid("parameter-selector-multi-resp")
 
-    expected_obs_responses = ["FOPR", "WOPR:OP1"]
+    dash_duo.wait_for_text_to_equal(
+        response_selector_id, "\n".join(expected_default_responses)
+    )
+
     obs_radio_btn = dash_duo.find_element(
         "#" + plugin.uuid("response-observations-check")
     )
     obs_radio_btn.click()
-    obs_responses = get_options(dash_duo, "#" + resp_selector_id)
-    assert obs_responses == expected_obs_responses
+
+    dash_duo.wait_for_text_to_equal(
+        response_selector_id, "\n".join(expected_responses_with_observations)
+    )


valentin-krasontovitsch:
jenkins, retest this please

valentin-krasontovitsch:
jenkins, retest this please

valentin-krasontovitsch:
ja so i tried in another PR as well (#361), but besides finding a different flaky spot, i was not able to reproduce it. so I suggest to merge these changes (after review), addressing both #297 and #360, and wait and see if it still fails :disappointed:

tests/views/test_response_correlation.py
```diff
@@ -118,16 +116,16 @@ def test_show_respo_with_obs(mock_data, dash_duo):
...
-    default_responses = get_options(dash_duo, "#" + resp_selector_id)
-
     expected_def_responses = ["FGPT", "FOPR", "SNAKE_OIL_GPR_DIFF", "WOPR:OP1"]
-    assert default_responses == expected_def_responses
-
     expected_obs_responses = ["FOPR", "WOPR:OP1"]
+
+    resp_selector_id = "#" + plugin.uuid("parameter-selector-multi-resp")
+
+    dash_duo.wait_for_text_to_equal(resp_selector_id, "\n".join(expected_def_responses))
```
hnformentin, line 124: Any reason to not having the timeout specification? `, timeout=5`

tests/views/test_response_correlation.py
```diff
@@ -118,16 +116,16 @@ def test_show_respo_with_obs(mock_data, dash_duo):
...
+
     obs_radio_btn = dash_duo.find_element(
         "#" + plugin.uuid("response-observations-check")
     )
     obs_radio_btn.click()
-    obs_responses = get_options(dash_duo, "#" + resp_selector_id)
-    assert obs_responses == expected_obs_responses
+
+    dash_duo.wait_for_text_to_equal(resp_selector_id, "\n".join(expected_obs_responses))
```
hnformentin, line 131: Any reason to not having the timeout specification? `, timeout=5`

tests/views/test_response_correlation.py
hnformentin: What means `def` in `expected_def_responses`? If it is default, wouldn|t it be better to have ` default` instead of `def`?

hnformentin:
Nice PR, we need to get good at using the wait_for wrappers and this a good start :clap: 
Out of scope here, but do you understand in which context we would prefer [wait_for_property_to_equal()](https://pytest-dash.readthedocs.io/en/latest/pytest_dash.html#pytest_dash.wait_for.wait_for_property_to_equal) in place of `wait_for_text_to_equal` ?

tests/views/test_response_correlation.py
valentin-krasontovitsch: definitely good to write out default, not a lot of reason anymore nowadays to use abbreviated words, in my humble opinion ^^

i didn't see a reason to deviate from the default timeout (10 secons I guess?) in this particular, so i went with the default.

do you see a reason to deviate from the default? is it too long? : )

tests/views/test_response_correlation.py
valentin-krasontovitsch: i like going with sane defaults, assuming that dash put a sane default : )

but if you think it's too much (is it 10 seconds, i don't remember), we can put our own value, but we probably have to do it all over... i dunno

tests/views/test_response_correlation.py
valentin-krasontovitsch: did a couple of renames expanding abbreviated words.

valentin-krasontovitsch:
It seems like `wait_for_property_to_equal` is for html tag properties (is that what they're called?), as in `<div x-data="foo">bar</div>` - for checking `bar`, we'd use `wait_for_text_to_equal`, I suppose, while we should use `wait_for_property_to_equal` for checking the value of `x-data`. I think. not sure though. didn't test it O:)

valentin-krasontovitsch:
> Nice PR

thank you ^^

tests/views/test_response_correlation.py
hnformentin: Maybe my question was misleadingly in the wrong place...I noticed that on other parts of the code a timeout was specified as 5 (e.g. lines 86 and 90), but not here. If there is no argument to add it here, is it an argument to remove the timeout of the others? 🤔

tests/views/test_response_correlation.py
hnformentin: Timeout comment replied below. Nice :-)

hnformentin:
> It seems like `wait_for_property_to_equal` is for html tag properties (is that what they're called?), as in `<div x-data="foo">bar</div>` - for checking `bar`, we'd use `wait_for_text_to_equal`, I suppose, while we should use `wait_for_property_to_equal` for checking the value of `x-data`. I think. not sure though. didn't test it O:)

Thanks! I was thinking that maybe we could check that we have the property ytitle tested, but thinking after your comment, it does not make sense since what we want to test is that the titles are correct and not that the title exists...Even if we tested ytitle, there are probably many elements with this property in the page. 

`<text class="ytitle" transform="rotate(-90,5.233203125000003,214.5)" style="font-family: &quot;Open Sans&quot;, verdana, arial, sans-serif; font-size: 14px; fill: rgb(42, 63, 95); opacity: 1; font-weight: normal; white-space: pre;" x="5.233203125000003" y="214.5" text-anchor="middle" data-unformatted="Value" data-math="N">Value</text>`

hnformentin (approved):

valentin-krasontovitsch:
yeah those elements are kinda difficult to test... thanks for approving!

tests/views/test_response_correlation.py
valentin-krasontovitsch: made issue #362

PR #363, opened 2022-08-15
Push implementation of service out of ert

eivindjahren:
In order to avoid install webviz_ert from ert,
the deprecated ert_hooks is removed and the vis service
executable from ert is moved to webviz_ert

This PR has to be coordinated with https://github.com/equinor/ert/pull/3757

## Pre review checklist


webviz_ert/__main__.py
```diff
@@ -0,0 +1,114 @@
...
+            if page["experimental"]:
+                return experimental_mode
+        return True
+
+    new_config_dict["pages"] = [
+        page
+        for page in new_config_dict["pages"]
+        if filter_experimental_pages(page, experimental_mode)
+    ]
+
```
hnformentin, line 61: Clean Code -> Vertical order -> function below a function that does the calling. Is here an exception?

hnformentin:
The error seems not to be a flakiness:

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
webviz-config 0.5.0a0 requires bleach[css]>=5; python_version >= "3.7", but you have bleach 4.1.0 which is incompatible.

ModuleNotFoundError: No module named 'webviz_ert.ert_hooks'
ERT crashed unexpectedly with "No module named 'webviz_ert.ert_hooks'".

eivindjahren:
> The error seems not to be a flakiness:
> 
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts. webviz-config 0.5.0a0 requires bleach[css]>=5; python_version >= "3.7", but you have bleach 4.1.0 which is incompatible.
> 
> ModuleNotFoundError: No module named 'webviz_ert.ert_hooks' ERT crashed unexpectedly with "No module named 'webviz_ert.ert_hooks'".

I thought I had already fixed that. I will have a closer look.

webviz_ert/__main__.py
eivindjahren: No, I simply move this function from ert to webviz_ert. I'll clean it up :)

eivindjahren:
Jenkins, test this please!

eivindjahren:
The failure on jenkins seems to be due to ert picking up the hook, even though we have removed it. I think this might be due to the bleeding environment this is being deployed into being out of sync. I suggest ignoring this for now and hopefully it will be fixed with rebuilding bleeding.

webviz_ert/__main__.py
eivindjahren: Fixed now.

pinkwah (approved):

PR #368, opened 2022-08-17
Fix obvious flakiness

### LINKED ISSUE #362: fix obvious potential flakiness
Cf. [meta-issue about flakiness](#359)

One way of testing that has (allegedly) introduced flakiness is grabbing an element with a dash method (which waits), and then inspecting its text or other property.

Instead, we should use [dash's provided methods](https://dash.plotly.com/testing) (~~or rather [`pytest-dash`'s](https://pytest-dash.readthedocs.io/en/latest/usage.html)~~ EDIT: pytest-dash is deprecated) for checking things directly, like `wait_for_text_to_equal`.

At the same time, we should decide on whether we should use the default timeout of dash always, or if there are instances where we want to set a custom timeout - there should be a reason for setting a custom timeout!

The goal of this issue is then to look through our existing tests, and find those spots where we can potentially reduce flakiness by using the methods mentioned above, and remove any unwarranted usage of a custom timeout.

origin of these findings is a discussion in [a PR about flakiness](#358)

valentin-krasontovitsch:
**Issue**
Closes #362

**Approach**
See issue description - gist is to use dash's methods for checking text and whatnot.

NB: following the camp-site rule, there's not only rewriting of tests to be less flaky, but also refactoring / renaming here and there, and fixing of little things (like removing clicking by coordinates, using existing helper methods to DRY up the code, correcting copypasta comments) - I hope that is acceptable, if not, mea culpa, I will minimize the changes.

In general I'm sorry, the changes are gonna be a long and possibly boring read, rest assured that a long (and a bit tedious, if i may say so) process created them, too :sweat_smile: 

## Pre review checklist

### DIFF
--- a/tests/views/test_ensemble_selector.py
+++ b/tests/views/test_ensemble_selector.py
@@ -12,44 +12,53 @@ def test_ensemble_refresh(
 ):
     plugin = setup_plugin(dash_duo, __name__, ParameterComparison)
 
-    ensemble_selector_id = plugin.uuid("ensemble-multi-selector")
-    initial_cases = get_options(dash_duo, "#" + ensemble_selector_id)
+    ensemble_selector_id = f"#{plugin.uuid('ensemble-multi-selector')}"
+    expected_initial_ensembles = [
+        "default",
+        "default_smoother_update",
+        "default3",
+        "nr_42",
+    ]
+    dash_duo.wait_for_text_to_equal(
+        ensemble_selector_id, "\n".join(expected_initial_ensembles)
+    )
 
     # Select first ensemble
     first_ensemble_name = select_ensemble(dash_duo, plugin)
 
     # Select a parameter
-    param_name = select_first(
+    first_parameter = select_first(
         dash_duo, "#" + plugin.uuid("parameter-selector-multi-param")
     )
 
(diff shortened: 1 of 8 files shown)


tests/views/test_ensemble_selector.py
```diff
@@ -12,44 +12,53 @@ def test_ensemble_refresh(
...
     # Select a parameter
-    param_name = select_first(
+    first_parameter = select_first(
         dash_duo, "#" + plugin.uuid("parameter-selector-multi-param")
     )
 
-    dash_duo.wait_for_contains_text(
-        "#" + plugin.uuid("parameter-deactivator-param"), f"×{param_name}", timeout=4
+    selected_parameters_finder = (
+        f"#{plugin.uuid('parameter-deactivator-param')} " ".Select-value-label"
```
oysteoh, line 35: do we actually have names ( css classes ) with a capital first letter or is this mispelled?

valentin-krasontovitsch: i think that's not us as in webviz-ert, it's the `Dropdown` component from `dash-core-components`, but yeah, we do:

css-class-names

oysteoh (approved):

PR #373, opened 2022-08-17
Fix index for clicking on observation

### LINKED ISSUE #369: Response Correlation Viewer: x index for observations misbehaves
**Describe the bug**
Clicking on a red dot representing an observation in the response plot does not, as expected, set the currently active x index to the index of the clicked observation, but rather sets it to (somewhere close to?) 0.

**To Reproduce**
3. Choose at least one ensembles, and a response with observations
4. Click on the red dot representing an observation
5. The vertical line indicating the index is set to somewhere far left

**Expected behavior**
The vertical line indicating the index is set to the index of the clicked observation

**Screenshots**
Having selected some index in the middle, before clicking on an observation:

obs-index-before

after clicking on the third from the left visible observation:

obs-index-after

 - OS: Ubuntu jammy 22.04
 - Browser: Google Chrome 104.0.5112.79 
 - Version: 79412d7bb1c81a177a52caef73de3319af3e3188

valentin-krasontovitsch:
Closes #369

**Approach**
We check whether the click was right on an observation, and do a reverse lookup of the clicked x value in the response axis - the previously used clicked index for an observation point refers to the observation axis, which is potentially shorter than the response axis.

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -17,6 +17,7 @@ from webviz_ert.plugins import WebvizErtPluginABC
 from webviz_ert.models import (
     load_ensemble,
     BarChartPlotModel,
+    EnsembleModel,
     PlotModel,
     Response,
 )
@@ -236,7 +237,11 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
 
             if response.observations:
                 for obs in response.observations:
-                    obs_plots.append(_get_observation_plots(obs.data_df()))
+                    observation_dataframe = obs.data_df()
+                    metadata = len(observation_dataframe) * ["observation"]
+                    obs_plots.append(
+                        _get_observation_plots(observation_dataframe, metadata)
+                    )
 
         fig = go.Figure()
         for plot in response_plots:
@@ -481,7 +486,12 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
                 )
         elif click_data:
             active_response = active_resp_param["response"]
-            corr_xindex[active_response] = click_data["points"][0]["pointIndex"]
+            x_index = _get_index_after_click_on_response_plot(
+                click_data["points"][0], active_response, loaded_ensemble
+            )
+            if x_index is None:
+                corr_xindex[active_response] = DEFAULT_X_INDEX
+            corr_xindex[active_response] = x_index
         return corr_xindex
 
     def _get_default_x_index(response: Response) -> int:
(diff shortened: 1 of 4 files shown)


valentin-krasontovitsch:
I've added a bunch of inline comments for this one, please feel free to take offense in that : )

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -703,3 +709,50 @@ def _add_plots_to_multiplot(
             row_histograms,
             2,
         )
+
+
+def _get_index_after_click_on_response_plot(
```
oysteoh, line 714: this seems to work ok - and i guess i could also approve with this. Just wanted to ask if you have looked into adding more metadata which could be queried on the datapoint? Is it possible to add i.e `type` to the point and make that value `observation` which then could be used instead of color? 

Another thing noticed here - not part of this fix - but maybe nice to have. Would it be an easy thing making the line "snap" to the closest observation? A dataseries with just a couple of observations you immediately now which observation you want to look at maybe, but without zooming in it can be hard to hit the exact spot where it is... :thinking:

valentin-krasontovitsch: that sounds like a great idea - i didn't want to dive into the code, as I wasn't sure how big the rabbit hole would be, but i will take a look, thanks for the pointer!

hmmmmmm yeah that's a good point, I also heard somebody make a comment like people are mostly interested in what happens at observations.
a possible implementation would be to see, if the response has observations, if the click was sufficiently close to one - one would have to define a threshold, and a distance function for x values (assuming numerical values and dates, this is no problem).
all in all, sounds doable.

and yet - we do have an observation selector in the pipeline, which one could argue would make the described request even more of a luxury.

then again, i'm really into the webviz-ert stuff right now, so I could probably quickly whip up a proof of concept, if you'd like me to.

what do you think?

valentin-krasontovitsch: rebased onto most recent merged work, and added metadata to observation traces in plot in question, and use that metadata to identify the points.

i'm just gonna open an issue about "clicking close to an observation", and copypaste our discussion there. one more aspect we should perhaps consider there is that users should still be able to choose a datapoint close but not equal to an observartion...

valentin-krasontovitsch: created #375 addressing the fuzzy observation selection.

oysteoh: > one more aspect we should perhaps consider there is that users should still be able to choose a datapoint close but not equal to an observartion...

good point :+1:  :thinking:

oysteoh (approved):

PR #384, opened 2022-08-23
Re observation index selector

### LINKED ISSUE #341: Scatter plot does not survive refresh
Before refresh:
Screenshot from 2022-07-28 18-31-02

after refresh:

Screenshot from 2022-07-28 18-32-18

Desired behaviour:

The Scatter plot should look the same after page refresh.

### LINKED ISSUE #342: Only one ensemble present in the heatmap plot
Selecting more then one ensemble does not show more then one ensemble in the heatmap plot:
Screenshot from 2022-07-28 18-53-21

### LINKED ISSUE #343: Scatter plot not updated when selecting second ensemble
Screenshot from 2022-07-28 19-04-51

In the image above two ensembles are selected the scatter plot only shows data from one, by clicking in the heatmap or adding/removing a parameter the scatter plot is updated. 

It should be updated also when an ensemble is selected/deselected

### LINKED ISSUE #345: Heatmap does not update selection of observations selector
**Describe the bug**
When two responses are selected, the observation selector displays two x-axis. When I move the one on the top to the left, it changes the values selected with the box, but the corresponding observations are not updated.

**To Reproduce**
See selections in the print screen.

**Expected behavior**
The observations in the heat map should correspond to the observations selected in the observation selector.

**Screenshots**
Observe how the heatmap is the same, even if the axis of the upper response is 800 in the screenshot from above and 1400 from the screenshot below.
Screenshot 2022-08-01 at 13 07 15
Screenshot 2022-08-01 at 13 23 11

### LINKED ISSUE #346: Bugs introduced by observation selector for heat map
**Describe the bug**
A series of bugs were introduced with #332 . We were aware bugs would be introduced, but we decided to merge the PR both to have early user feedback and allow to solve the issues together.

**Issues detected**

- [x] #340 
- [ ] #341 
- [ ] #342  
- [ ] #343   
- [ ] #345 
- [ ] #347 
- [ ] #348
- [x] #349
- [ ] #350
- [ ] #352

### LINKED ISSUE #347: Legend of response in the observation selector not complete
**Describe the bug**
See selections in the screenshot. I would expect that the name of the response in the legend of the observation selector was the same as the one in the response selector. In the screenshot below, "SNAKE_OIL_" is being omitted.

**Screenshots**
Screenshot 2022-08-01 at 13 07 15

### LINKED ISSUE #348: x-axis value of the observations selected in the observation selector do not match heat map index
**Describe the bug**
The description of the responses in the heat map are appended with an index, which does not correspond necessarily to the stamp of the observations.

**To Reproduce**
See selections in the screenshot.

**Expected behavior**
I would expect to be able to see the time of the observation or how it is related to the response.
Screenshot 2022-08-01 at 13 07 15

### LINKED ISSUE #352: Response Correlation Viewer: Tornado and heatmap not drawn without selected observations
**Describe the bug**
Choosing some data on the response correlation viewer, the tornado and heatmap plot are not always drawn.

**To Reproduce**
3. Choose at least one ensemble, one response without observations and at least one parameter
4. The heatmap and tornado plots are not drawn

**Expected behavior**
The heatmap and tornado plots should be drawn

**Suspected explanation**
No default index indicated by a vertical black line in the response overview plot is chosen when the response is added to selection of responses. Without a chosen index, the relevant plots cannot be plotted (as they are a function of the index). When one on the other hand chooses a response with observations, and draws a rectangle for choosing observations in the observation overview plot, the tornado and heatmap plots are drawn.

DanSava:
**Issue**
Resolves #346
Resolves #341
Resolves #342
Resolves #343
Resolves #345 
Resolves #347
Resolves #348 
Resolves #352

### DIFF
--- a/tests/data/spe1_st/tests/test_webviz_ert.py
+++ b/tests/data/spe1_st/tests/test_webviz_ert.py
@@ -10,6 +10,7 @@ from tests.conftest import select_first, setup_plugin, verify_key_in_dropdown
 
 parameter_keys = ["FIELD_PROPERTIES:POROSITY", "FIELD_PROPERTIES:X_MID_PERMEABILITY"]
 response_keys = ["WGPT:PROD", "WWPT:PROD", "WOPT:PROD", "WWIT:INJ"]
+response_keys_with_observations = ["WOPT:PROD"]
 
 
 def _verify_keys_in_menu(dash_duo_handle, plugin, keys, selector):
@@ -55,7 +56,10 @@ def test_webviz_response_correlation(dash_duo):
     )
 
     _verify_keys_in_menu(
-        dash_duo, plugin, response_keys, "parameter-selector-multi-resp"
+        dash_duo,
+        plugin,
+        response_keys_with_observations,
+        "parameter-selector-multi-resp",
     )
     response_name = select_first(
         dash_duo, "#" + plugin.uuid("parameter-selector-multi-resp")
@@ -63,12 +67,13 @@ def test_webviz_response_correlation(dash_duo):
     param_name = select_first(
         dash_duo, "#" + plugin.uuid("parameter-selector-multi-param")
     )
+
     dash_duo.wait_for_text_to_equal(
         "#" + plugin.uuid("info-text"),
         "".join(
             [
                 f"RESPONSE: {response_name}",
-                f"INDEX: 2015-02-01",
+                f"INDEX: 2016-01-01",
                 f"PARAMETER: {param_name}",
             ]
         ),
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -1,26 +1,21 @@
 import pandas as pd
+from pandas._libs.tslibs.timestamps import Timestamp
 import plotly.graph_objects as go
 import dash
-import datetime
 import numpy as np
 import dash_bootstrap_components as dbc
 
 from typing import List, Optional, Dict, Tuple, Any, Union
 from copy import deepcopy
-from dash import dcc, html
+from dash import html
 from dash.development.base_component import Component
 from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output, State
 from plotly.subplots import make_subplots
+from plotly.graph_objs.layout import XAxis
 
 from webviz_ert.plugins import WebvizErtPluginABC
-from webviz_ert.models import (
-    load_ensemble,
-    BarChartPlotModel,
-    EnsembleModel,
-    PlotModel,
-    Response,
-)
+from webviz_ert.models import load_ensemble, BarChartPlotModel, PlotModel, AxisType
 from webviz_ert.controllers.multi_response_controller import (
     _get_observation_plots,
     axis_label_for_ensemble_response,
@@ -29,8 +24,6 @@ from webviz_ert import assets
 
 
 def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) -> None:
-    DEFAULT_X_INDEX = 0
-
     @app.callback(
         [
             Output(
@@ -44,7 +37,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         ],
         [
             Input(parent.uuid("ensemble-selection-store"), "data"),
-            Input(parent.uuid("correlation-store-xindex"), "data"),
+            Input(parent.uuid("correlation-store-selected-obs"), "data"),
             Input(parent.uuid("correlation-store-active-resp-param"), "data"),
             Input(parent.uuid("correlation-metric"), "value"),
             Input(parent.uuid("sort-parameters"), "on"),
@@ -57,7 +50,7 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
     )
     def update_correlation_plot(
         ensemble_selection_store: Dict[str, List],
-        corr_xindex: Dict,
+        selected_obs: Dict,
         active_resp_param: Dict,
         correlation_metric: str,
         sort_parameters: bool,
@@ -68,13 +61,18 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
         ensembles = _get_selected_ensembles_from_store(ensemble_selection_store)
         if not (
             ensembles
+            and selected_obs
             and parameters
             and responses
-            and active_resp_param["response"] in responses
+            and active_resp_param["response"]["name"] in responses
         ):
-            return ({}, {})
+            return {}, {}
 
         active_response = active_resp_param["response"]
+        active_response_name = active_response["name"]
+        active_index = active_response.get("index")
+        if active_index:
+            active_response_name = f"{active_response_name}::{active_index}"
         data = {}
         colors = {}
         heatmaps = []
@@ -86,20 +84,33 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             if parameter_df.empty:
                 continue
             valid_responses = []
-            for response in responses:
-                response_df = ensemble.responses[response].data_df()
-                if response_df.empty:
+            for response_name in responses:
+                response = ensemble.responses[response_name]
+                if response.empty:
                     continue
-                x_index = corr_xindex.get(response, 0)
-                parameter_df[response] = response_df.iloc[x_index]
-                valid_responses.append(response)
+                resp_indexes = []
+                valid_response_with_index = []
+                if response.observations:
+                    for obs_idx in selected_obs.get(response.name, []):
+                        if isinstance(obs_idx, str):
+                            resp_index = str(Timestamp(obs_idx))
+                        else:
+                            resp_index = obs_idx
+                        resp_indexes.append(resp_index)
+                        valid_response_with_index.append(f"{response.name}::{obs_idx}")
+                    valid_responses += valid_response_with_index
+                    new_dataframe = response.data.loc[resp_indexes].T
+                    new_dataframe.columns = valid_response_with_index
+                    parameter_df = pd.concat([parameter_df, new_dataframe], axis=1)
 
             corrdf = parameter_df.corr(method=correlation_metric)
             corrdf = corrdf.drop(valid_responses, axis=0).fillna(0)
-            if active_response not in valid_responses:
+            if active_response_name not in valid_responses:
                 continue
             if sort_parameters:
-                corrdf, df_index = _sort_dataframe(corrdf, df_index, active_response)
+                corrdf, df_index = _sort_dataframe(
+                    corrdf, df_index, active_response_name
+                )
             # create heatmap
             data_heatmap = {
                 "type": "heatmap",
@@ -116,8 +127,8 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             }
             heatmaps.append(data_heatmap)
             ens_key = repr(ensemble)
-            data[ens_key] = corrdf[active_response]
-            data[ens_key].index.name = active_response
+            data[ens_key] = corrdf[active_response_name]
+            data[ens_key].index.name = active_response_name
             colors[ens_key] = assets.get_color(index=index)
         if data and heatmaps:
             correlation_plot = BarChartPlotModel(data, colors)
@@ -130,12 +141,18 @@ def response_correlation_controller(parent: WebvizErtPluginABC, app: dash.Dash) 
             for idx, heatmap in enumerate(heatmaps):
                 heatmap_plot.add_trace(heatmap, 1, 1 + idx)
                 heatmap_plot.update_yaxes(showticklabels=False, row=1, col=1 + idx)
-                heatmap_plot.update_xaxes(showticklabels=False, row=1, col=1 + idx)
+                heatmap_plot.update_xaxes(
+                    showticklabels=True,
+                    tickangle=90,
+                    ticklabelposition="outside left",
+                    row=1,
+                    col=1 + idx,
+                )
                 if hide_hover:
                     heatmap_plot.update_traces(patch={"hoverinfo": "none"})
                 else:
                     heatmap_plot.update_traces(
-                        patch={"hovertemplate": "%{y}: %{z}<extra></extra>"}
+                        patch={"hovertemplate": "%{x} - %{y}: %{z}<extra></extra>"}
                     )
             _layout = assets.ERTSTYLE["figure"]["layout"].copy()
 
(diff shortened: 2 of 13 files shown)


DanSava:
The resulting changes look as follows:

response_correlation

oysteoh:
Think it looks great! But - i struggle to play with it - nothing happens when i'm clicking / selecting currently. Any clues? 🤔

oysteoh:
Could this change also include updates to the tutorial ?  And - i think this looks good and ready for feedback. Not everything is clear and make sense in the beginning, but i like it after  playing around. Easier to select the data you care about :)

DanSava:
> Could this change also include updates to the tutorial?

What tutorial are you referring to? 

> Easier to select the data you care about :)

Currently, only one selection rectangle is allowed but with a bit more work it is possible to allow multiple selection rectangles. But I would suggest  that as a future improvement and not be part of the current POC

oysteoh:
> What tutorial are you referring to?
The guided tour... I didn't notice it was changed and the new feature was documented ! Well done 👍

valentin-krasontovitsch:
having one ensemble, one response with a couple observartions, and at least one parameter, when selecting only one observation, and changing which observation is selected by drawing a rectangle around a new one, the correlation x index is not updated in the "active info" line on top, and also not in the scatterplot.

only clicking on the heatmap updates the index and the scatterplot.

is this a bug? should i open an issue for this?

valentin-krasontovitsch:
I see that handling of `corr_x_index` was merged back into `update_active_resp_param`, the docstring of `update_corr_index` removed, and the docstring of `update_active_resp_param` not updated - bummer that we lost that refactoring to separate handling of those things. i can see that the current handling of the `corr_x_index` is simpler, but it also seems be lacking, considering the above bug(?).

perhaps you could consider making it into its own function again, and describing how and when it gets updated.

valentin-krasontovitsch:
the heatmap info text should perhaps be updated to reflect that the heatmap can be used to change the active x index.

valentin-krasontovitsch:
we should also consider renaming `corr_x_index` to something like `active_observation_index` to reflect that we are really only interested in observations, not random points (and their indexes)

valentin-krasontovitsch:
i guess it's intentional that only responses with observations are visible in the response correlation viewer now? should that be documented / indicated somewhere?

valentin-krasontovitsch:
it looks like collapsing the selectors also leads to the observation selector being partly hidden - don't know if this is just me though...

collapse

valentin-krasontovitsch:
another little thing - when sliding the axes of the observation selector, the points under the selection rectangle move, but the selection doesn't get updated. i suppose this is something that can be fixed lated (should i open an issue for that?)

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -388,13 +535,14 @@ def update_info_text(
...
-        x_axis = ensemble.responses[active_response].axis
-        formatted_active_index = _format_index_value(
-            x_axis, corr_xindex.get(active_response, 0)
-        )
-        active_index_text = str(formatted_active_index)
+
+        selected_obs_indexes = selected_obs.get(active_response, None)
+        active_index_text = "NA"
+        if selected_obs and selected_obs_indexes:
+            active_index = corr_xindex.get(active_response) or selected_obs_indexes[0]
```
valentin-krasontovitsch, line 543: I think this piece of logic should be moved to the function that outputs `corr_xindex` - the default value, when no observation is chosen, for the active index, is the first observation.

the same logic is repeated in line 495 f.

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -388,13 +535,14 @@ def update_info_text(
...
         info_text = []
-        ensemble = load_ensemble(parent, ensembles[0])
         active_response = active_resp_param["response"]
-        x_axis = ensemble.responses[active_response].axis
-        formatted_active_index = _format_index_value(
-            x_axis, corr_xindex.get(active_response, 0)
-        )
-        active_index_text = str(formatted_active_index)
+
+        selected_obs_indexes = selected_obs.get(active_response, None)
```
valentin-krasontovitsch, line 540: this variable name should be singular, right? it's just one index?
and perhaps renamed to active_observation_index?

tests/data/spe1_st/tests/test_webviz_ert.py
```diff
@@ -55,20 +55,21 @@ def test_webviz_response_correlation(dash_duo):
     )
 
     _verify_keys_in_menu(
-        dash_duo, plugin, response_keys, "parameter-selector-multi-resp"
+        dash_duo, plugin, ["WOPT:PROD"], "parameter-selector-multi-resp"
```
valentin-krasontovitsch, line 58: i suggest making a new variable, `responses_with_observations`, to make it clearer why this test needed touching up

DanSava:
> having one ensemble, one response with a couple observartions, and at least one parameter, when selecting only one observation, and changing which observation is selected by drawing a rectangle around a new one, the correlation x index is not updated in the "active info" line on top, and also not in the scatterplot.
> 
> only clicking on the heatmap updates the index and the scatterplot.
> 
> is this a bug? should i open an issue for this?

It's a bug should be fixed

DanSava:
> I see that handling of `corr_x_index` was merged back into `update_active_resp_param`, the docstring of `update_corr_index` removed, and the docstring of `update_active_resp_param` not updated - bummer that we lost that refactoring to separate handling of those things.

I read somewhere that extensive function comments are the promise of future lies. I will update the docsctring for `update_active_resp_param` to match what is going on there now. 

> i can see that the current handling of the `corr_x_index` is simpler, but it also seems be lacking, considering the above bug(?).
> 
> perhaps you could consider making it into its own function again, and describing how and when it gets updated.

Not really a good idea to have it separate, more code to read, more code to understand, more code to maintain, to read more comments to understand and keep up to date with changes.

DanSava:
> the heatmap info text should perhaps be updated to reflect that the heatmap can be used to change the active x index.

Not sure what heatmap info you are referring to

valentin-krasontovitsch:
> Not sure what heatmap info you are referring to

the one in the guided tour : )

DanSava:
> i guess it's intentional that only responses with observations are visible in the response correlation viewer now? 

Yes 

> should that be documented / indicated somewhere?

not sure, are you referring to documentation for the user or for the developer?

valentin-krasontovitsch:
> the promise of future lies

yeah i can see how they demand a lot of discipline: have to be read and perhaps updated every time a function is touched...

> Not really a good idea to have it separate, more code to read, more code to understand, more code to maintain, to read more comments to understand and keep up to date with changes.

i guess we have very different opinions in this case - i think when i want to figure out how a variable is manipulated, then having a function that is dedicated to just that variable (and not a bunch of variables) makes it easier to read and understand what is going on.
also i thought we were trying to avoid returns with several values?

if the functions are clearly structured and named, and the variables are named clearly, precisely, and in a descriptive fashion, then one can perhaps forego the docstring. I've heard several people on the team say that dash and its callback functions are difficult to understand, and it sure as heck was hard for me to figure out what was going on... so yeah the docstrings increase maintenance costs, for sure. my hope is that they also make it easier to get into the code, and start doing changes and verifying behaviour - especially since we don't have any place where the expected behaviour of the app is documented!

yes, after remembering all this, i'm a very very strong advocate for documenting the callback functions with describing the expected behaviour of the app, as in what user interactions should trigger what changes in the views. the technical /implementation details we can skip.

valentin-krasontovitsch:
> not sure, are you referring to documentation for the user or for the developer?

not sure either - perhaps a place where it's obvious for both? maybe the label text of the response selector?

DanSava:
> another little thing - when sliding the axes of the observation selector, the points under the selection rectangle move, but the selection doesn't get updated. i suppose this is something that can be fixed lated (should i open an issue for that?)

I guess this is only for the top axis when there is both an index and a time axis in the same plot and you are sliding the top axis then the points are moving but the rectangle is fixed, correct?

> (should i open an issue for that?)

I think it is a good idea to have an issue to disable the sliding of axis.

valentin-krasontovitsch:
> I guess this is only for the top axis when there is both an index and a time axis in the same plot and you are sliding the top axis then the points are moving but the rectangle is fixed, correct?

yes, you're right! funny, didn't realize that.

gonna make an issue for ... disabling the sliding when the top and bottom axis are visible?

DanSava:
> > Not sure what heatmap info you are referring to
> 
> the one in the guided tour : )

what is not clear about this following ?

Heatmap based representation of correlation (-1, 1) among all selected responses and parameters, for currently selected ensembles in column-wise fashion.One can select a currently active response and parameter by clicking directly on a heatmap. 

> One can select a currently active response and parameter by clicking directly on a heatmap. 

I think it is pretty self-explanatory that you can click on the heatmap and that new active response and parameter will be selected

valentin-krasontovitsch:
> what is not clear about the following?

oh no, it's clear, don't get me wrong. sorry, i should have been more verbose - i would add a comment about choosing the active observation (index), as it's the only place where one can jump between chosen observations

DanSava:
> yes, you're right! funny, didn't realize that.
> 
> gonna make an issue for ... disabling the sliding when the top and bottom axis are visible?

I would just disable sliding it should not matter if there are both axis or not

valentin-krasontovitsch:
> I would just disable sliding it should not matter if there are both axis or not

coolio!

DanSava:
> yeah i can see how they demand a lot of discipline: have to be read and perhaps updated every time a function is touched...
> 
> > Not really a good idea to have it separate, more code to read, more code to understand, more code to maintain, to read more comments to understand and keep up to date with changes.
> 
> i guess we have very different opinions in this case - i think when i want to figure out how a variable is manipulated, then having a function that is dedicated to just that variable (and not a bunch of variables) makes it easier to read and understand what is going on. also i thought we were trying to avoid returns with several values?

Yes, we have very different opinions in this case. 

> also i thought we were trying to avoid returns with several values?

That is very true when you write your own logic, but Dash is built with the premise that you should return multiple outputs at the same time, they actually encourage this ,not really sure why, probably because you cannot return the same output from multiple callbacks also might be related to callback order when you have the same input triggering different outputs.  

> 
> if the functions are clearly structured and named, and the variables are named clearly, precisely, and in a descriptive fashion, 

This is very subjective...

I've heard several people on the team say that dash and its callback functions are difficult to understand, and it sure as heck was hard for me to figure out what was going on... so yeah the docstrings increase maintenance costs, for sure. my hope is that they also make it easier to get into the code, and start doing changes and verifying behaviour - especially since we don't have any place where the expected behaviour of the app is documented!

I agree, extensively documenting a page which is marked as experimental just seems excessive given that it will be changed very frequently in the future.

webviz_ert/controllers/response_correlation_controller.py
DanSava: Very good suggestion!

valentin-krasontovitsch: : )

valentin-krasontovitsch:
> Dash is built with the premise that you should return multiple outputs at the same time, they actually encourage this

I see! alrighty then.

> extensively documenting a page which is marked as experimental just seems excessive given that it will be changed very frequently in the future.

that's good point : )

webviz_ert/controllers/response_correlation_controller.py
```diff
@@ -152,25 +162,116 @@ def update_correlation_plot(
...
+                opacity=0.35,
+            )
+        else:
+            # Use first index if nothing is selected
+            for p in obs_plots:
+                selected_indexes[p.name] = [p.x_axis[0]]
+
+        parent.save_state("correlation-store-obs-range", store_obs_range)
+        return (
+            {"min-height": "auto", "padding-top": "20px"},
```
valentin-krasontovitsch, line 269: i think this one might be responsible for the collapse of the observation selector plot when top selectors (ensembles, responses, parameters) get collapsed - it ain't pretty, i don't like hard coding sizes, but i also have a very bad handle on css in general, but setting this to `300px` stopped it from collapsing on me

valentin-krasontovitsch, line 269: i think this one might be responsible for the collapse of the observation selector plot when top selectors (ensembles, responses, parameters) get collapsed - it ain't pretty, i don't like hard coding sizes, but i also have a very bad handle on css in general, but setting this to 300px stopped it from collapsing on me

valentin-krasontovitsch (approved):
looks very good, sorry about the chaotic review yesterday (not using the review feature), i didn't expect that i would have a lot of comments...

just to summarize here what i'm still missing:

- collapsing observation selector plot (see own comment about size)
- adjust guided tour info text for heatmap to include that, besides active response and parameter, the currently active observation can be changed by clicking on the heatmap
- rename `corr_x_index` to something like `active_observation_index`

i'll leave it up to you what of the above you want to include in this PR

webviz_ert/controllers/response_correlation_controller.py
DanSava: Perfect, that was very helpful :+1:

DanSava: I refactor some more of the code yesterday and integrated the concept of `corr_x_index` or at least some part of the concept of `corr_x_index` into the `active_resp_param`

valentin-krasontovitsch: "hiding" the index in the active response - elegant, i like it : ) then one doesn't need to think so much about it, as its own thing.

hnformentin:
Sorry for coming late to the party. It took a while to fix  my environment which was broken...

Some observations, not sure if this is the intended behaviour:

1. If I have a response selected with multiple observations selected in the selector, like here:

Screenshot 2022-08-30 at 10 31 15

and I select another response, then I loose track of the observations selected previously:

Screenshot 2022-08-30 at 10 31 30


2. It is not possible to select a time or observation from the response plot (e.g. move the vertical red bar to select a different point).

3. In the observation selector, the axis are movable to the right and the left. Nevertheless, if there is an interval selected and I move the axis, it does not update the selection box.
For example, here I had one index selected and two timestamps:
Screenshot 2022-08-30 at 10 39 48

But when I moves the index axis (and bring two observations in to the box), the data displayed is not updated. 
Screenshot 2022-08-30 at 10 40 08

I think that would be more sensible that the axis was not movable or the data get updated when we move the axis.

valentin-krasontovitsch:
@hnformentin sorry it's hidden in that wall of text:

- the axis moving is noted, and an issue has been created to disable it
- the selection of a current index / observation is indeed not possible through clicking on the response overview plot anymore, one must / can only click on the heatmap

hnformentin:
I can continue playing a lottle bit, please let me know if I should add the comments elsewhere...

DanSava:
@hnformentin Thank you for taking the time to review 

> Sorry for coming late to the party. It took a while to fix my environment which was broken...
> 
> Some observations, not sure if this is the intended behaviour:
> 
> 1. If I have a response selected with multiple observations selected in the selector,  and I select another response, then I loose track of the observations selected previously:

Yes, that is the intended behavior, what you are doing there is selecting a response with observations that are time indexed and then selecting a response with observations that are value indexed. In this case the observation selection was made only for time index. That is why the selection range is reset  when a new response is added and the first observation for each of the responses is considered active. 

> 2. It is not possible to select a time or observation from the response plot (e.g. move the vertical red bar to select a different point).

True, this is by design, only by clicking the heat-map will the selected observation index change. 
 
> 3. In the observation selector, the axis are movable to the right and the left. Nevertheless, if there is an interval selected and I move the axis, it does not update the selection box.
> I think that would be more sensible that the axis was not movable or the data get updated when we move the axis.

as @valentin-krasontovitsch already stated this will be addressed in a further issue and the axis moving should be disabled.

hnformentin:
Nice, it seems you have all in control 👏 👏 👏 
Thanks @valentin-krasontovitsch and @DanSava for addressing the comments...

PR #389, opened 2022-08-31
Force indexes for responses and response observations to be consistent

DanSava:
**Issue**
Resolves bleeding issues :crossed_fingers: 

**Approach**

* Force indexes for responses and response observations to be of type pandas.Timestamp
* Also, avoid using python automatic bool conversion


oysteoh (approved):
LGTM! Made all of my interaction work without any exceptions now - seems like the issues with indexes has been resolved :)

PR #396, opened 2022-09-19
Remove hack poking storage to get test to pass

### LINKED ISSUE #168: Fix integration komodo test in order not to poke the storage first
Currently the test `test_webviz_parameter_comparison` `tests/data/spe1_st/tests/test_webviz_ert.py` requires to poke the storage first (ie. via `get_ensemble_id`) in order that the gui elements are populated correctly. We should try to find out why is that and remove it.

oysteoh:
Removing old hack to get test to run - apparently it was enough to just remove it.

Resolves #168

### DIFF
--- a/tests/data/spe1_st/tests/conftest.py
+++ b/tests/data/spe1_st/tests/conftest.py
@@ -1,38 +1,10 @@
-import requests
 import pytest
 
 
-@pytest.fixture(scope="session")
-def get_info():
+@pytest.fixture(scope="session", autouse=True)
+def start_storage_server():
     from ert.shared.services import Storage
 
     with Storage.start_server() as service:
-        yield service.fetch_url(), service.fetch_auth()[1]
-
-
-@pytest.fixture(scope="session")
-def requests_get(get_info):
-    base_url, token = get_info
-
-    def req_get(url, headers=None, params=None):
-        if headers is None:
-            headers = {}
-        resp = requests.get(
-            f"{base_url}/{url}",
-            headers={**headers, "Token": token},
-            params=params,
-        )
-        return resp
-
-    return req_get
-
-
-@pytest.fixture(scope="session")
-def get_experiment_dict(requests_get):
-    experiments = requests_get("experiments").json()
-    return experiments[0]
-
-
-@pytest.fixture(scope="session")
-def get_ensemble_id(get_experiment_dict):
-    return get_experiment_dict["ensemble_ids"][0]
+        service.wait_until_ready(timeout=30)
+        yield service
--- a/tests/data/spe1_st/tests/test_webviz_ert.py
+++ b/tests/data/spe1_st/tests/test_webviz_ert.py
@@ -29,9 +29,7 @@ def _verify_keys_in_menu(dash_duo_handle, plugin, keys, selector):
 
 
 @pytest.mark.spe1
-def test_webviz_parameter_comparison(get_ensemble_id, dash_duo):
-    # here we need the poke storage first - to get this test running
-    _ = get_ensemble_id
+def test_webviz_parameter_comparison(dash_duo):
     plugin = setup_plugin(dash_duo, __name__, ParameterComparison)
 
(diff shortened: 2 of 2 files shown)


tests/data/spe1_st/tests/test_webviz_ert.py
```diff
@@ -30,8 +30,6 @@ def _verify_keys_in_menu(dash_duo_handle, plugin, keys, selector):
 
 @pytest.mark.spe1
 def test_webviz_parameter_comparison(get_ensemble_id, dash_duo):
```
DanSava, line 32: Looks good, but the actual call happens when the fixture `get_ensemble_id` is initialized, I think if the fixture is removed from the definition of the test, all the tests will from the file will fail

oysteoh: Ok - i'll have a look :+1:

oysteoh: Allright - i've refactored a bit to make things more clear. The problem here is that without the call to `get_ensemble_id` the fixture for starting the storage server was never launched.

oysteoh: Utilizing autouse and session scope, forcing the server to startup before tests are ran. In addition i changed start_server to connect_or_start_server to avoid multiple calls to start_server - and then just return the url and token ( used in request_get fixture ) . This seems like a more clean approach to me.

tests/data/spe1_st/tests/conftest.py
```diff
@@ -2,17 +2,17 @@
...
-    with Storage.start_server() as service:
+    with Storage.connect_or_start_server() as service:
         yield service.fetch_url(), service.fetch_auth()[1]
 
 
-@pytest.fixture(scope="session")
-def requests_get(get_info):
-    base_url, token = get_info
+@pytest.fixture()
+def requests_get(start_storage_server):
```
DanSava, line 14: I might be wrong here but I think after the refactor you can remove this fixture and also the `get_experiment_dict` fixture

tests/data/spe1_st/tests/conftest.py
```diff
@@ -2,17 +2,17 @@
...
 
 
-@pytest.fixture(scope="session")
-def get_info():
+@pytest.fixture(scope="session", autouse=True)
+def start_storage_server():
     from ert.shared.services import Storage
 
-    with Storage.start_server() as service:
+    with Storage.connect_or_start_server() as service:
```
DanSava, line 9: Why change `start_server` with `connect_or_start_server` ?

tests/data/spe1_st/tests/conftest.py
```diff
@@ -2,17 +2,17 @@
 import pytest
 
 
-@pytest.fixture(scope="session")
-def get_info():
+@pytest.fixture(scope="session", autouse=True)
+def start_storage_server():
```
DanSava, line 6: I would refactor it a bit:

```python
@pytest.fixture(scope="session", autouse=True)
def start_storage_server():
    from ert.shared.services import Storage
    with Storage.start_server() as service:
        service.wait_until_ready(timeout=20)
        yield service
    service.shutdown()
```

DanSava (approved):
I think it looks good :+1: had just some possible improvement suggestions

tests/data/spe1_st/tests/test_webviz_ert.py
DanSava: Using  `scope="session"` should call the fixture only once per session so technically there is no need for the replacement.

tests/data/spe1_st/tests/conftest.py
oysteoh: Shouldn't the service shutdown when the context closes ? Is it necessary to have one explicitly doing it?

DanSava: That's right it is not needed

PR #398, opened 2022-09-21
Update location of ert services

eivindjahren:
The location of ert services has moved in ert=4.0.0-b0. This updates the path which is used only in the spe1 integration test. It seems `webviz-ert` runs the spe1 integration test on whatever is deployed in komodo bleeding, so the timing of this merge depends on the state of what is deployed.

## Pre review checklist


DanSava (approved):
Looks good :+1:

PR #405, opened 2023-01-24
Update chromedriver supporting chromium 108

oysteoh:
**Issue**
Testing of GUI requires chromedrivers supporting the chromium installation to be installed in the test-environment. Chromium is now updated to version 108.xx and needs a new chromedriver.


valentin-krasontovitsch (approved):
nice catch!

PR #415, opened 2023-02-20
Install chromedriver to match installed chromium

### LINKED ISSUE #414: Guarantee chromium version for integration test
We currently use Chrome for the integration test in `ci/jenkins/testkomodo.sh`. 

Accordingly, we download the `chromedriver` here

https://github.com/equinor/webviz-ert/blob/main/ci/jenkins/testkomodo.sh#L23

Currently, this specifies v108, but I need `chromedriver` version 109.0.5414.74 for tests to pass on RGS nodes on the Azure server s034 (where `chromium-browser --version` is `Chromium 109.0.5414.119 Fedora Project`). I don't know if/when we'll need other versions in other testing environments.

:question: Should we download Chromium as part of the test setup, so we know what we're getting? This seems like it would avoid the possibility of having to debug failing tests due to driver mismatch in the future (because the error is a bit obscure).

kwinkunks:
Instead of having to update the chromedriver to match the browser whenever the test breaks, we can ask Google's chromedriver API for the correct version dynamically. This should let us cope with the possibility of having to run tests on different platforms, with potentially different versions of Chrome/Chromium.

**Issue**
Resolves #414

**Approach**
Rather than installing chromium ourselves (which would be slow), or using a 3rd party lirbary to detect or install the right version (which adds a dependency), and taking advantage of [Google's API](https://chromedriver.chromium.org/downloads/version-selection) for this, we add a couple of extra steps to `testkomodo.sh`.

## Pre review checklist

### DIFF
--- a/ci/jenkins/testkomodo.sh
+++ b/ci/jenkins/testkomodo.sh
@@ -19,8 +19,10 @@ start_tests () {
 
 start_integration_test () {
 
-    echo "Downloading chromedriver ..."
-    wget -O chromedriver.zip https://chromedriver.storage.googleapis.com/108.0.5359.71/chromedriver_linux64.zip
+    chromium_version=$(chromium-browser --version | grep -zoP '\d+\.\d+\.\d+')
+    chromedriver_version=$(curl -s https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$chromium_version)
+    echo "Downloading chromedriver v$chromedriver_version for chromium-browser v$chromium_version"
+    wget -O chromedriver.zip https://chromedriver.storage.googleapis.com/$chromedriver_version/chromedriver_linux64.zip
     unzip chromedriver.zip chromedriver -d ../test-kenv/root/bin
 
     pip install pytest selenium dash[testing]


ertomatic:

ci/jenkins/testkomodo.sh
```diff
@@ -19,8 +19,10 @@ start_tests () {
 
 start_integration_test () {
 
-    echo "Downloading chromedriver ..."
-    wget -O chromedriver.zip https://chromedriver.storage.googleapis.com/108.0.5359.71/chromedriver_linux64.zip
+    chromium_version=`chromium-browser --version | grep -zoP '\d+\.\d+\.\d+'`
+    chromedriver_version=`curl -s https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$chromium_version`
```
berland, line 23: Should we use dollar-parens form instead of backticks?  http://mywiki.wooledge.org/BashFAQ/082

kwinkunks: Doh, agreed.

berland (approved):
:+1:

PR #416, opened 2023-02-21
unset HOST env var before integration test

### LINKED ISSUE #411: Integration tests fail if $HOST is set
This is about the tests in ``tests/data/spe1_st/tests`.

The environment variable `HOST` is used by the `dash` server:

https://github.com/plotly/dash/blob/ae093af869ad6a57292b35a065c8c953bbb90444/dash/dash.py#L1822

However, Selenium expects the server to be `localhost`, so if `HOST` is set to something other than `localhost` or `127.0.0.1` (e.g. to the name of the host machine, which was the case for some testing on Azure) then Selenium will experience a `ERR_CONNECTION_REFUSED` error:

```
ERROR    dash.testing.application_runners:application_runners.py:197 expected condition not met within timeout
Traceback (most recent call last):
  File "/prog/komodo/2023.02.b4-py38-rhel7/root/lib/python3.8/site-packages/dash/testing/application_runners.py", line 192, in start
    wait.until(
  File "/prog/komodo/2023.02.b4-py38-rhel7/root/lib/python3.8/site-packages/dash/testing/wait.py", line 26, in until
    raise TestingTimeoutError(msg)
dash.testing.errors.TestingTimeoutError: expected condition not met within timeout
DEBUG    selenium.webdriver.remote.remote_connection:remote_connection.py:361 POST http://localhost:58124/session/467f1131ce40f661833a0c6ebcf3649c/url {"url": "http://localhost:58052"}
DEBUG    urllib3.connectionpool:connectionpool.py:456 http://localhost:58124 "POST /session/467f1131ce40f661833a0c6ebcf3649c/url HTTP/1.1" 500 1054
DEBUG    selenium.webdriver.remote.remote_connection:remote_connection.py:383 Remote response: status=500 | data={"value":{"error":"unknown error","message":"unknown error: net::ERR_CONNECTION_REFUSED
```

In order to help notice this in the future, it might be a good idea to assert that `HOST` is not set before running the tests marked `spe1`. Perhaps this would be best done in `tests/data/spe1_st/tests/conftest.py`.

Possibly `HOST` should be unset in `ci/jenkins/testkomodo.sh`, but for now we're taking care of it in the `komodo-releases` workflow file, https://github.com/equinor/komodo-releases/blob/main/.github/workflows/base-unit-test.yml

[Issue in `komodo-releases`](https://github.com/equinor/komodo-releases/issues/3191)

kwinkunks:
Only the f_komodo user seems to have  set in in bash. when the Dash server starts it picks up  for the server name, but Selenium is looking for 127.0.0.1 or localhost.

**Issue**
Resolves #411

**Approach**
Unset the environment variable before running the tests.

## Pre review checklist

### DIFF
--- a/ci/jenkins/testkomodo.sh
+++ b/ci/jenkins/testkomodo.sh
@@ -30,12 +30,15 @@ start_integration_test () {
     pushd tests/data/spe1_st
 
     echo "Initiating Ert run for Spe1 with new storage enabled..."
-
     ert ensemble_experiment --enable-new-storage spe1.ert
-
     echo "Ert ensemble_experiment run finished"
 
-    echo "Test for webviz-ert plugins ..."
+    # $HOST is set to machine name for f_komodo user's bash shells.
+    # Must unset for Dash to use localhost as server name, as Selenium expects.
+    unset HOST
+
+    echo "Test for webviz-ert plugins..."
     pytest ../../../ -vs -m spe1
+
     popd
 }


ertomatic:

berland (approved):
Nice catch

PR #417, opened 2023-02-22
Make license consistent in PyPI and README

### LINKED ISSUE #410: Match README badge to actual license, or add to metadata make dynamic
The license on the project is GPL, based on the [`LICENSE`](https://github.com/equinor/webviz-ert/blob/main/LICENSE) file.

However, the README contains the badge for the AGPL. Additionally, the license is not named in the project metadata in `setup.py` or anywhere else.

Since we publish the package to PyPI, if we included the licence in the metadata we could pick it up from there with, e.g. 

```
[![PyPI license](https://img.shields.io/pypi/l/webviz-ert.svg)](https://pypi.org/project/webviz-ert/)
```

Bottom line: we should at least change the badge to match the license (assuming we're keeping the GPL, which I'd recommend).

kwinkunks:
The wrong license was hardcoded into the README badge. This update instead gets it from PyPI, which in turn uses the project metadata. Since the metadata was not set, I added it to setup.py -- so it won't appear in PyPI until the next release. After that, all will be good.

I also took the opportunity to add a bit more metadata.

**Issue**
Resolves #410

**Approach**
Use PyPI metadata for badge.

## Pre review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![PyPI version](https://badge.fury.io/py/webviz-ert.svg)](https://badge.fury.io/py/webviz-ert)
 [![Build Status](https://github.com/equinor/webviz-ert/workflows/Python/badge.svg)](https://github.com/equinor/webviz-ert/actions?query=workflow%3APython)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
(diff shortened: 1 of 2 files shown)


ertomatic:

berland (approved):
nice!

PR #419, opened 2023-02-28
Fix flaky test on self-hosted runner

### LINKED ISSUE #418: Fix flaky test on self-hosted runner
On Azure, the test `test_response_correlation.test_axes_labels()` is flaky, timing out half the time with the default 10 s limit. [Here is one example](https://github.com/equinor/komodo-releases/actions/runs/4292680318/jobs/7481029809):

```
=================================== FAILURES ===================================
_______________________________ test_axes_labels _______________________________

mock_data = None
dash_duo = <dash.testing.composite.DashComposite object at 0x2b4017208970>

    def test_axes_labels(mock_data, dash_duo):
        """test_axis_labels loads two different plots and checks that axes are
        labelled correctly"""
    
        plugin = setup_plugin(
            dash_duo, __name__, ResponseCorrelation, window_size=(1024, 2048)
        )
    
        # find the right ensemble which has mock data prepared for this test
        wanted_ensemble_name = "default3"
        select_ensemble(dash_duo, plugin, wanted_ensemble_name)
    
        wanted_responses = ["WOPR:OP1", "FOPR"]
    
        # we only see one response at a time, so we choose and check one after the
        # other
        for wanted_response in wanted_responses:
            response_selector_id = plugin.uuid("parameter-selector-multi-resp")
            select_by_name(dash_duo, f"#{response_selector_id}", wanted_response)
    
            plot_id = plugin.uuid("response-overview")
    
            # check that y axis label spells out "Value"
>           dash_duo.wait_for_text_to_equal(f"#{plot_id} text.ytitle", "Value")

tests/views/test_response_correlation.py:34: 
.
.
.
E       selenium.common.exceptions.TimeoutException: Message: text -> Value not found within 10s

../test-kenv/root/lib64/python3.8/site-packages/selenium/webdriver/support/wait.py:87: TimeoutException
```

Following tests depend on the same plot. Perhaps it takes about 10 s to render (seems long but :shrug: ) so propose increasing timeout limit for this one wait.

kwinkunks:
One of the Selenium tests on a plot axis title was sometimes timing out. This change doubles the timeout to 20s. For sure, 10s already seems long; if there's still an issue it may be another problem.

**Issue**
Resolves #418

**Approach**
Double the length of the timeout.

## Pre review checklist

### DIFF
--- a/tests/views/test_response_correlation.py
+++ b/tests/views/test_response_correlation.py
@@ -30,8 +30,9 @@ def test_axes_labels(mock_data, dash_duo):
 
         plot_id = plugin.uuid("response-overview")
 
-        # check that y axis label spells out "Value"
-        dash_duo.wait_for_text_to_equal(f"#{plot_id} text.ytitle", "Value")
+        # check that y axis label spells out "Value"; test is flaky so longer
+        # timeout.
+        dash_duo.wait_for_text_to_equal(f"#{plot_id} text.ytitle", "Value", timeout=20)
 
         # check that one has date, the other has index as x axis label
         if wanted_response == "FOPR":


ertomatic:

berland:
The other `dash_duo.wait_for_text_to_equal()` in the same file, should they have a specific timeout as well, or is it only this one that has triggered failures?

kwinkunks:
@berland So far I've only seen the test fail at this line. My impression is that after `select_by_name()` there is only one plot, and each `wait_for_text_to_equal()` is just looking at it. So I thought that perhaps only the first one might need to wait a bit longer for the plot to render.

berland (approved):
Lets give it a try!

Issue #426, opened 2023-03-29
Fix flaky selenium tests


kwinkunks:
Seems we are having more than just 1 flaky test that may need a longer timeout in order to pass on Azure.

Already fixed #418 with PR #419.

New problems will reference this one, which should stay open for the Azure testing and pilot.

kwinkunks:
Closed #427 which addresses more of this.

PR #427, opened 2023-03-29
More flaky timeouts

kwinkunks:
**Issue**
Does not resolve but works on #426, which we're leaving open for more fixes if required.

**Approach**
We previously increased the timeout limit on similar tests in the same file; this commit doubles the timeout in the same way for the other functions.

## Pre review checklist

- [x] (I don't think there are any, maybe we should add Cloud or similar) ~Added appropriate labels~

### DIFF
--- a/tests/views/test_response_correlation.py
+++ b/tests/views/test_response_correlation.py
@@ -30,8 +30,9 @@ def test_axes_labels(mock_data, dash_duo):
 
         plot_id = plugin.uuid("response-overview")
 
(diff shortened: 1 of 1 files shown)


ertomatic:

kwinkunks:
May be obviated by https://github.com/equinor/webviz-ert/pull/424

xjules (approved):
I think it's good @kwinkunks to have it just in case, so :rocket: !

Issue #428, opened 2023-04-07
webviz-ert: adjust timeout


xjules:
We do have another flaky test, which fails when testing in Azure:
```sh
FAILED tests/views/test_parameter_selector.py::test_search_input_return_functionality - selenium.common.exceptions.TimeoutException: Message: text -> default not found inside element within 4s
```

xjules:
Closing this one as being part of #426

PR #429, opened 2023-04-14
Simplify dash timeout

kwinkunks:
**Issue**
Timeouts were set in a few different places, and we're trying to change them and add others to cope with Azure performance... but it's getting out of hand. Does not resolve but is connected to the perma-issue #428.

**Approach**
- Make a way to set the explicit and implicit timeouts in one place: `conftest.py`
- Make a way to set the timeouts from the command-line invocation of pytest
- Double the timeouts when the `testkomodo.sh` script detects the `$CI_RUNNER_LABEL` env var has the value`azure` 
- Remove all the custom timeout overrides
- Remove a slightly hacky `wait_a_bit` function which added custom sleeps
- Remove a bunch of commented-out code

## Pre review checklist


ertomatic:

kwinkunks:
Three tests timed out, so will increase the timeout in the workflow and see if that fixes it.

Also Selenium seems to be throwing a lot of warnings, non-fatal.

kwinkunks:
Doubled both timeouts (implicit and explicit) and two tests are failing. I don't tihnk the yare flaky / performance related, and they failed on the last PR before this one too, but I can't see what the issue is right now. **I cannot reproduce it locally.**

hnformentin:
As tests are failing, I would think it is not ready for review. I will move to todo, sorry if it was a mistake from my side.

PR #438, opened 2023-08-14
Download driver correctly for chromium-browser v115+

### LINKED ISSUE #437: Process to download ChromeDriver is broken
Tests on Azure are failing: https://github.com/equinor/komodo-releases/actions/runs/5844553665/job/15851778533

`wget` fails to download the ChromeDriver: `key does not 'exist.</Message><Details>No' such object: 'chromedriver/LATEST_RELEASE_115.0.5790</Details></Error>/chromedriver_linux64.zip'`

Reason: as of Chromium 115 (12 July), which has just been introduced on our servers, the process to download ChromeDriver hs changed. See https://chromedriver.chromium.org/downloads/version-selection

Sounds like it should be a bit easier now:

> As a result, you might no longer have a need for version selection — you could choose any available CfT version and simply download the correspondingly-versioned ChromeDriver binary. For automated version downloading one can use the convenient [CfT JSON endpoints](https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints).

I guess until onprem catches up to v115, we'll need to be able to do both, by testing for version >= 115, or similar.

kwinkunks:
**Issue**
Resolves #437

**Approach**
Check the major version to decide which download method to use.

We should be able to simplify again soon, once SVG RGS nodes upgrade `chromium-browser` and there's no chance of downgrading again. Will make an issue to check and simplify.

## Pre review checklist


xjules (approved):
Looks very good! :rocket:

PR #443, opened 2023-10-16
Pin urllib3 < 2

oysteoh:
urllib3 needs to be pinned below v2 due to incompabilites with RHEL7

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,16 +3,6 @@ from setuptools import setup, find_packages
 with open("README.md", "r") as fh:
     LONG_DESCRIPTION = fh.read()
 
-TESTS_REQUIRE = [
-    "mypy",
-    "bandit",
-    "black",
-    "dash[testing]",
-    "pylint",
-    "pytest-mock",
-    "selenium~=3.141",
-]
-
 setup(
     name="webviz-ert",
     description="Webviz plugins for ERT",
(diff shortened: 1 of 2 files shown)


berland:
FYI, in the github action workflow that tests webviz-ert, urllib3 is already pinned, but `pip` still installs a newer urllib3.[

https://github.com/equinor/komodo-releases/blob/e39c27012e15aeea78d863a2c236a5aacf89fd61/.github/workflows/run_tests_one_project.yml#L102

oysteoh:
> FYI, in the github action workflow that tests webviz-ert, urllib3 is already pinned, but `pip` still installs a newer urllib3.[
> 
> https://github.com/equinor/komodo-releases/blob/e39c27012e15aeea78d863a2c236a5aacf89fd61/.github/workflows/run_tests_one_project.yml#L102

I guess it is because the new version is installed through a dependency when installing webviz-ert, and by pinning it this should now be solved? Or am i thinking wrong here?

oyvindeide:
Some test dependency must install a newer version then?

oysteoh:
> Some test dependency must install a newer version then?

i.e selenium uses requests which uses urllib3 - so i guess it is pulled down when running the tests. Updated the PR to add the pinning there instead.

oyvindeide (approved):

PR #447, opened 2023-11-28
Remove old chromedriver download method

### LINKED ISSUE #439: Simplify chromedriver download in `testkomodo.sh`
PR #438 made the chromedriver download process in `testkomodo.sh` much more complicated. Once we have `chromium-browser --version` showing v117 or so (i.e. 115 plus a bit) on all platforms, we can revert to a single source (i.e. downloading from https://edgedl.me.gvt1.com).

kwinkunks:
This leaves only the new download method, since v115. Resolves #439.


berland (approved):
Looks good, assuming this has been tested.

PR #458, opened 2024-01-10
Allow curl to fail when file 404s

### LINKED ISSUE #457: Chrome driver downloads still failing
The fix did not work -- the 'test' download of headers fails the job: https://github.com/equinor/komodo-releases/actions/runs/7468565343/job/20325358434

Need another way to check if the download exists. Or could fall back on the lowest-valued release for this minor version. First will check again that Google does not have another way to do this.

kwinkunks:
I guess we run with set -e and curl returns 1 when the file is 404. So this version always returns true from the curl, and we don't exit. Fixes #457.

**Issue**
Resolves #457

**Approach**
Main thing was adding the `|| true` to the `curl --head` that tests the download.

Also allowing other versions of HTTP as I bet it's actually 1.1 -- so even the 'file exists' case might have failed before in fact.

## Pre review checklist

### DIFF
--- a/ci/testkomodo.sh
+++ b/ci/testkomodo.sh
@@ -19,13 +19,13 @@ start_tests () {
 
 start_integration_test () {
 
-    chromium_version=$(chromium-browser --version | grep -zoP '\d+\.\d+\.\d+\.\d+')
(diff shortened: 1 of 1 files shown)


sondreso (approved):
LGTM!

PR #461, opened 2024-01-19
Remove unused install function in testkomodo.sh

larsevj:
I can not see that this is in use, and this function would potentially reinstall the entire package. When testing webviz-ert the dependencies is instead installed via the default install_test_dependencies in komodo/func_def.sh.


kwinkunks (approved):
Indeed, is no longer used here or by https://github.com/equinor/komodo/blob/main/ci/func_def.sh

Good catch :+1:

PR #462, opened 2024-02-19
Update the download URL for the chrome driver

larsevj:
It seems like google has again changed the download URLs of the chrome driver. This aims to remedy this. Must be accompanied by a corresponding firewall change.


kwinkunks (approved):
Looks reasonable, and cleaner than before. Seems to work nicely.

Hopefully this does it for a while!

PR #466, opened 2024-06-27
Mark browser-tests for conditional executions

### LINKED ISSUE #446: Mark tests requiring dash / Selenium so they can be turned off
The `chromium-browser` version on Azure at the moment does not have a `chromedriver` available, and trying to download it throws a 404: https://github.com/equinor/komodo-releases/actions/runs/6885078281/job/18729226670

This should never happen, it seems to be a bug on RedHat's part (?).

AFAIK, there's no other way to get Chrome drivers -- we are following the advice here https://googlechromelabs.github.io/chrome-for-testing/ -- some versions simply do no exist.

But we could mark tests requiring the driver, however, so that if situations like this happen again, we can easily turn off the tests temporarily.

AugustoMagalhaes:
**Issue**
Resolves #446

**Approach**
Following the issue's comment made by @kwinkunks
 "But we could mark tests requiring the driver, however, so that if situations like this happen again, we can easily turn off the tests temporarily.", the approach went on marking the tests that requires dash / Selenium as "browser-test", which can be skipped with the command "pytest --skip-browser-tests". 

## Pre review checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,18 +8,6 @@ from selenium import webdriver
 from tests.data.snake_oil_data import ensembles_response
 
 
-def validate_chromedriver():
-    try:
-        driver = webdriver.Chrome()
-        browser_version = driver.capabilities["browserVersion"]
-        chromedriver_version = driver.capabilities["chrome"][
-            "chromedriverVersion"
-        ].split(" ")[0]
-        return browser_version[0:3] == chromedriver_version[0:3]
-    except (WebDriverException, TypeError):
-        return False
-
-
 def pytest_addoption(parser):
     parser.addoption(
         "--skip-browser-tests",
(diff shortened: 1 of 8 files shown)


eivindjahren:
Hi! Thanks for your contribution! We are looking at your PR now, but we are currently coming back from vacation. Don't be afraid to reach out if you have any questions :)

tests/conftest.py
```diff
@@ -2,12 +2,51 @@
...
+    )
+
+
+def pytest_collection_modifyitems(config, items):
+    skip_browser_tests = pytest.mark.skip(
+        reason="chromedriver missing in PATH or intentionally skipped"
+    )
+    valid_chromedriver = validate_chromedriver()
+    browser_tests = [item for item in items if "browser_test" in item.keywords]
+    if valid_chromedriver and not config.getoption("--skip-browser-tests"):
```
eivindjahren, line 44: We probably want the case where the chromedriver version is invalid and `--skip-browser-test` is not given to produce an error message. This is so that tests are not silently skipped unintentionally.

eivindjahren (changes requested):
The PR overall looks really good! I just had a small comment about the behavior in case of invalid chromium driver version.

tests/conftest.py
AugustoMagalhaes: Hello Eivind, thanks for the comment, I've just pushed the changes. Have a nice day! :)

eivindjahren (approved):

PR #471, opened 2024-08-08
Handle ensemble schema without parent and child ensemble_ids

DanSava:
**Issue**

The ensemble schema sometimes it does not contain parent and child ensemble_ids.


andreas-el (approved):
Mighty fine sir! 🎖️

PR #472, opened 2024-08-12
Disable observation index selector xaxis and yaxis panning

### LINKED ISSUE #386: Response Correlation Viewer: Disable sliding in observation selector
**Describe the bug**
The observation selector allows for displaying observations from several responses. When responses with different axis (units) are displayed, the sliding functionality of the plot for the upper axis does not interact well with the selection rectangle.

**To Reproduce**
1. Open up webviz-ert with data that containts at least two responses with observations with different axes (e.g. in `ert`, `test-data/local/snake_oil`)
3. Choose one ensemble and two responses with axes with different units (e.g. `FOPR` and `SNAKE_OIL_WPR_DIFF@199`
4. Draw a rectangle to select a couple of observations (at least one)
5. Draw the upper axis of the observation selector plot left or right
6. Observe that the upper observation points move, while the selection rectangle, the lower points, and the observation selection remain constant
7. Likewise, when dragging the lower axis, the rectangle and the lower points move, but the upper points remain stationary, and the selection does not change.

**Expected behavior**
For example both axes and all points and the rectangle always move, and the selection remains constant, or whenever only parts move, the selection is updated to reflect the new visual situation.

We suggest disabling the sliding of the axes in the observation selector plot completely, cf. (bits of a discussion on the observation selector PR)[https://github.com/equinor/webviz-ert/pull/384#issuecomment-1230216478]

DanSava:
**Issue**
Resolves #386 

**Approach**
Fix range for observation index selector for xaxis and yaxis. 

## Pre review checklist

### DIFF
--- a/webviz_ert/controllers/response_correlation_controller.py
+++ b/webviz_ert/controllers/response_correlation_controller.py
@@ -11,7 +11,7 @@ from dash.development.base_component import Component
 from dash.exceptions import PreventUpdate
 from dash.dependencies import Input, Output, State
 from plotly.subplots import make_subplots
-from plotly.graph_objs.layout import XAxis
+from plotly.graph_objs.layout import XAxis, YAxis
 
 from webviz_ert.plugins import WebvizErtPluginABC
 from webviz_ert.models import load_ensemble, BarChartPlotModel, PlotModel, AxisType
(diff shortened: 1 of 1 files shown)


eivindjahren (approved):
:bike:

PR #481, opened 2025-02-13
Fix numpy warning

larsevj:
Fixups warnings. Tests will still install old urllib3 due to constraints with selenium and wekzeugs apparently, but seems difficult if not impossible to fix at the moment.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -5,6 +5,6 @@ black
 scipy
 dash[testing]>=2.5.1
 selenium
-urllib3<2 # RHEL7 does not support urllib >= 2.0
+urllib3<2 # Pinned to make Snyk happy. Also indirectly pinned due to restrictions in other test deps.
 flask>=2.2.5 # not directly required, pinned by Snyk to avoid a vulnerability
 werkzeug>=3.0.6 # not directly required, pinned by Snyk to avoid a vulnerability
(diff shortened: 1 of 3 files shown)


.github/workflows/python.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     - name: Test with pytest
       run: |
         pip install -r test_requirements.txt
-        pytest
+        pytest -v
```
jonathan-eq, line 35: Do we want it verbose?

test_requirements.txt
```diff
@@ -5,6 +5,6 @@ black
 scipy
 dash[testing]>=2.5.1
 selenium
-urllib3<2 # RHEL7 does not support urllib >= 2.0
+urllib3<2 # Pinned to make Snyk happy. Also indirectly pinned due to restrictions in other test deps.
```
jonathan-eq, line 8: Does snyk have issues with urrllib3 >= 2?

jonathan-eq: Do you think they will be fixing the vulnerability pointed out by Snyk, or is it not really maintained?

jonathan-eq: I guess we will keep it then. Great finding!

jonathan-eq (approved):
Very good!

PR #504, opened 2026-01-07
Silence the pkg_resource warnings to avoid noise

andreas-el:
Silence the pkg_resource warnings to avoid noise when running from ert

## Pre review checklist


yngve-sk (approved):
Nice, works for me!🚀

